PR #5, opened 2015-05-14
Fix float/double error w/ unpack (tuple)

rodcarroll:
- Added test case to verify doubles/floats
- Added gitignore for the git folks


tebeka:
Thanks

PR #9, opened 2015-07-20
No .decode on str type when no codec is given.

oliverbestmann:


tebeka:
Thanks! Will try to merge soon-ish (back from vacation nuking my inbox).

PR #10, opened 2015-07-23
Basestring does not exist in python3. Use str instead.

DoHe:
In python3 basestring does not exist and str should be used instead.
Compare string_types https://pythonhosted.org/six/#constants.
Parallel to what was done before I added this into the existing six.py file.

### DIFF
--- a/fastavro/six.py
+++ b/fastavro/six.py
@@ -26,6 +26,8 @@ if sys.version_info >= (3, 0):
     def py3_json_dump(obj, indent):
         json.dump(obj, stdout, indent=indent)
 
+    basestring = str
+
 else:  # Python 2x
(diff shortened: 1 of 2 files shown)


DerDot:
As far as I can see, the failing tests are not connected to this pull request.
Before this pull request, the run-tests script already died on the flake8 step, which is not recognized by travis as failing (probably because of the way $? is used).

pkoch:
I think I ended up fixing this on #11. However, my code is a little bit more encompassing: I will accept `bytes` as strings, as I'm assuming you already have them encoded as UTF-8. This way, we sidestep making people box their data in `unicode` objects.

DoHe:
Yes, seems you did. I think your pull request is preferable over this one, so I'm gonna close it. Thanks.

PR #15, opened 2015-08-05
Add enclosing namespace to unqualified references

dcreager:
When you refer to a previously defined schema using an unqualified name, and there's an enclosing `namespace` to work with, we need to [join the two together](http://avro.apache.org/docs/current/spec.html#Names) to get the fully qualified name to look for — unless the unqualified name is one of the primitive types, in which case it should stay unqualified.


tebeka:
Thanks!

PR #16, opened 2015-08-11
Allow nested record definitions from arrays and maps.

mtth:
Also fix namespace bug for `union` references. (Added a test for each.)

Thanks for writing this great package!


tebeka:
Thanks! Will check shortly and merge.

wesm:
Why are the compiled files being checked in?

rouge8:
:+1: this fixes some bugs I was going to track down

mtth:
@wesm - Didn't look into it much but the README has this snippet:

> As recommended by Cython, the C files output is distributed. This has the advantage that the end user does not need to have Cython installed. However it means that every time you change fastavro/pyfastavro.py you need to run make.

wesm:
The recommendation is for setup.py sdist, not git. See pandas and other modules implemented in large part in Cython

virtuald:
+1

tebeka:
Merged. Will leave the generated c files out of git.

Issue #17, opened 2015-08-13
Writing and reading sets boolean to False


DoHe:
I use following schema:

```
{'fields': [{'name': 'flag', 'type': 'boolean'}],
 'name': 'test',
 'type': 'record'}
```

to write this datum:

```
{"flag":True}
```

Serializing results in this byte string:

```
b'Obj\x01\x04\x14avro.codec\x08null\x16avro.schema\xa6\x01{"name": "test", "type": "record", "fields": [{"type": "boolean", "name": "flag"}]}\x00:\xea\xf7\x8cV\xa0:4\\\xd0A\x85\xa4\x05\xcf\xe9\x02\x021:\xea\xf7\x8cV\xa0:4\\\xd0A\x85\xa4\x05\xcf\xe9'
```

and deserializing it gives me this datum:

```
[{'flag': False}]
```

i.e. the True has been turned to False. The deserialized schema is intact.

Additionally, writing a datum like `{}` works just fine (although it should throw an Exception) and when reading back, I again get:

```
[{'flag': False}]
```

I'm using python3, but have reproduced the problem on python2 as well. I also tried the pure python version without the compiled cython extensions and have the same problem there. I use a BytesIO as the fo, if this matters.

tebeka:
Thanks for reporting! Will try to have a look soon-ish.

ahane:
an avro file containing just a True value (which fails when added to the test files)
https://github.com/ahane/fastavro/blob/booleans/tests/avro-files/bool.avro

DoHe:
If I serialize with the reference implementation and deserialize with fastavro, everything works just fine, so the problems seems to be in the serialization already.

scottbelden:
I put a PR in to fix this. Located here: https://github.com/tebeka/fastavro/pull/18

PR #18, opened 2015-08-14
fix boolean encoding

scottbelden:
Resolves https://github.com/tebeka/fastavro/issues/17

The reader checks for ord(fo.read(1)) == 1, however a boolean is written using fo.write(b'1') or fo.write(b'0'). Using chr(1) and chr(0) instead like in the normal avro library works.

```python
>>> ord(b'1')
49
>>> ord(b'0')
48
>>> ord(chr(1))
1
>>> ord(chr(0))
0
```

EDIT: Oops. chr() was taken out for python34 compatibility. Therefore we should just check for the correct value (49)


tebeka:
Thanks! Will merge soon-ish.

PR #19, opened 2015-08-18
handle records with no fields

scottbelden:
Currently if you try to do the following:

```python
schema = { 
    'name': 'Test',
    'namespace': 'test',
    'type': 'record',
    'fields': [
    ],  
}

records = [ 
    {}, 
]

with open('fastavrotest.avro', 'wb') as out:
    writer(out, schema, records)
```

Then when you read the container you will get back an empty list of records. I've included an avro file that fails currently:

```python
AssertionError: assert [] == [{}]
```

But passes with the fix.


tebeka:
Thanks! I'll merge soon-ish...

tebeka:
Out in 0.8.8, thanks!

PR #20, opened 2015-08-18
handle error types

scottbelden:
The reader could already read error types but the writer was unable to serialize them.


tebeka:
Thanks! Will merge soon-ish. (You're fast man :)

scottbelden:
We've been using the standard avro library for a while and are looking to switch to this as it is indeed much faster. As a result we have schemas for almost everything and so just trying to get this to handle all of them.

tebeka:
Are there any more in the pipeline? I'll hold of the release. If not will try to release early tomorrow (GMT+2)

scottbelden:
Possibly. Also, I think I accidentally added an extra commit to this PR. Let me fix that.

PR #23, opened 2015-08-18
fix issue resolving ints and longs in unions

scottbelden:
Last one for the time being, but this one is divergent from the official avro library handling (which has the same problem) so you can choose to do with it what you will (but I plan to keep it in my version either way).

Assume you have a field in a schema that is a union between an int and a long. If you have a container where that field was 123L and encoded as a long, when it gets deserialized and re-serialized by the Python code, it will now be encoded as an int. This isn't really a problem when you only have python users, but issues can (and have for us) come up when the other endpoint is something like Java. What happens is that Java properly encodes the value as a long, but when it roundtrips through the python code it comes back to Java encoded as an int. The same comment was made at the end of this message in the mailing list: http://mail-archives.apache.org/mod_mbox/avro-user/201304.mbox/%3cCA+i_aEkzOdL1d6WvZAPbm5VmrgD_8UE-8HC25Rf7WagijjJiMA@mail.gmail.com%3e

### DIFF
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -54,6 +54,10 @@ def read_boolean(fo, schema):
     return ord(fo.read(1)) == 49
 
 
+def read_int(fo, schema):
+    return int(read_long(fo, schema))
+
+
 def read_long(fo, schema):
     '''int and long values are written using variable-length, zig-zag
     coding.'''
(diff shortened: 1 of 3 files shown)


scottbelden:
Ah, as you can probably tell now, I still live in the 2.7 world. Looks like this type of thing won't work for 3.x unless there is some more sophisticated type hinting. Going to close this then.

PR #25, opened 2015-08-20
write the default value if the record has one

scottbelden:
If a field has a default value it's nice to have the writer go ahead and use that value when serializing (instead of throwing an error about the field missing).


tebeka:
Thanks!

PR #26, opened 2015-08-20
correct boolean roundtrip

scottbelden:
Hopefully no more boolean PRs... the previous PR was setting all booleans to True. Copying the associated test case and running it on master shows that.

Previously:

```python
In [22]: pack('B', 1) != 0x00
Out[22]: True

In [23]: pack('B', 0) != 0x00
Out[23]: True
```

Now:

```python
In [25]: unpack('B', pack('B', 1))[0] != 0
Out[25]: True

In [26]: unpack('B', pack('B', 0))[0] != 0
Out[26]: False
```

### DIFF
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -54,7 +54,7 @@ def read_boolean(fo, schema):
 
     # technically 0x01 == true and 0x00 == false, but many languages will cast
     # anything other than 0 to True and only 0 to False
-    return fo.read(1) != 0x00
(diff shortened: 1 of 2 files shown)


fastavro/reader.py
```diff
@@ -54,7 +54,7 @@ def read_boolean(fo, schema):
 
     # technically 0x01 == true and 0x00 == false, but many languages will cast
```
tebeka: Maybe we'd like to be more strict here and allow only 0 and 1? Will think about it.

PR #29, opened 2015-08-28
Fix writer.validate to look for bytes rather than str for a fixed or bytes field, ensuring it works with python 3

mjwillson:
In python2 `str is bytes` but in python 3 `str` won't do for fixed/bytes fields (and infact reader decodes as `bytes`)

### DIFF
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -156,7 +156,7 @@ def validate(datum, schema):
         return is_str(datum)
 
     if record_type == 'bytes':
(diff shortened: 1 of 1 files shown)


tebeka:
Thanks for the PR, will merge soon-ish.

Is it possible to add a test (that fails on the current version with Python 3)?

mjwillson:
Yeah sorry about lack of test. I'm afraid this was a bit of a yak shave while trying to get something else working, but I'll try and return to it if I get a moment.

PR #30, opened 2015-08-31
only populate the necessary repo

scottbelden:
Instead of populating both the READERS and WRITERS repo on any operation, only the relevant repo will be populated. For a very complex schema I noticed a 5-10% speed up in writing and a 10-20% speed up in reading.


tebeka:
Thanks

PR #31, opened 2015-09-03
add support for schema migration

scottbelden:


tebeka:
Thanks. Sorry, little busy - will try to get to it soon-ish....

fastavro/reader.py
```diff
@@ -42,12 +42,33 @@
...
 
-def read_null(fo, schema):
+class SchemaResolutionError(Exception):
+    pass
+
+
+def match_types(writer_type, reader_type):
+    # The writer and reader types might not be avro primitives, they could be
+    # named records or enums
+    if isinstance(writer_type, list) or isinstance(reader_type, list):
```
tebeka: Don't you mean `and` (instead of `or`)?

scottbelden: This was intended. The reason is that if the writer was an array of ints and the reader was an array of a union of ints or strings, then that should be no problem and the resolution will happen in read_union.

However, when adding some tests to show this, I found an issue. Namely, if the writer schema is a boolean or other primitive type, and the reader schema is something else, a SchemaResolutionError isn't thrown because the check isn't done.

scottbelden:
I added some more tests for some union cases that I realized were broken. I tried to add some comments where it seemed necessary, but questions are welcome. I don't believe any of this should slow down the normal reading much when there is no new updated schema as it should just add only a few "if reader_schema" checks.

tebeka:
@scottbelden Can you merge/rebase master?

scottbelden:
Rebased on top of master. Although, in the process I realized if there is a way to parameterize test functions that would be useful as many of the tests that were added are the same functions with just different schemas and record data. I haven't had a chance to look into how to do this with nose.

PR #34, opened 2015-09-15
always overwrite repo values

scottbelden:
If a record name was already defined in the repo it would not be overwritten by a new record which would cause reads/writes to potentially fail (since they would use the original schema added to the repo)

### DIFF
--- a/fastavro/schema.py
+++ b/fastavro/schema.py
@@ -67,7 +67,7 @@ def extract_named_schemas_into_repo(schema, repo, transformer, parent_ns=None):
 
     namespace, name = schema_name(schema, parent_ns)
 
-    if name and (name not in repo):
+    if name:
         repo[name] = transformer(schema)
 
     schema_type = schema.get('type')
--- a/tests/test_fastavro.py
+++ b/tests/test_fastavro.py
@@ -248,3 +248,63 @@ def test_metadata():
     new_file.seek(0)
     new_reader = fastavro.reader(new_file)
     assert new_reader.metadata['key'] == metadata['key']
+
+
+def test_repo_caching_issue():
+    schema = {
+        "type": "record",
+        "name": "B",
+        "fields": [{
+            "name": "b",
+            "type": {
+                "type": "record",
+                "name": "C",
+                "fields": [{
+                    "name": "c",
+                    "type": "string"
+                }]
+            }
+        }]
+    }
+
+    new_file = MemoryIO()
+    records = [{"b": {"c": "test"}}]
+    fastavro.writer(new_file, schema, records)
+    new_file.seek(0)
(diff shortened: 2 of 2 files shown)


fastavro/schema.py
```diff
@@ -67,7 +67,7 @@ def extract_named_schemas_into_repo(schema, repo, transformer, parent_ns=None):
 
     namespace, name = schema_name(schema, parent_ns)
 
-    if name and (name not in repo):
```
tebeka: This was added in the initial write support and probably for a reason :)
Is this how the spec says it should be?

scottbelden: I can't find anything in the spec. Perhaps @pkoch can shed some light as he was the one who originally added it. If I had to guess it was probably just some micro optimization about not reading/writing to the cache dict if the schema was already in there. But that's precisely the problem. A second record with the same name could come along and when reads/writes are done, the first schema that was added is used. The test case I added fails for me under the current master.

tebeka:
Merged, thanks!

pkoch:
Ah! I was expecting that, and I took the opposite stance from you.

The purpose of this is to *not* let you acquaint with new schemas. If you re-read the old schema, I won't bother to process it since I already know that name.

The current implementation is not really spec compliant in any way. We should have different repositories of schemas so we can work with more than one schema set.

scottbelden:
@pkoch I don't quite follow. Especially if you are doing read operations, how would you know that the schemas you are reading belong to different sets? In the test case I attached with this PR, there are two schemas. If you had a serialized file of each those schemas on hand and read the file using the smaller schema first and then read the file using the larger schema, the second file would be read incorrectly.

Why would you not deserialize with the schema embedded in the binary?

pkoch:
I wrote this from a perspective of writing records to Kafka topics. My typical use case would be to get acquainted with the full schema once in the initialization and emit single serialized records. As you can expect, it's common practice to not embed the schema on every message (hence my difference in perspectives). If I happened to mistakenly re-acquaint with the schema, it would in my best interest to do the least amount of work possible. That's where that came from.

Your approach is much sounder when thinking in files. Acquainting with the embedded schema is pretty much mandatory and should be done for every file, read or written. However, bear in mind it is inherently flawed for multi-threaded environments (as you'll be mutating global shared state). Even besides that, I cannot underline enough the need for separate schema repos. This starts getting really hairy once you start thinking about having support for reader schemas and schema evolution in general. But that's out of scope here.

Looking at this now, I've switched sides to your approach. Go forth and serialize/deserialize!

tebeka:
Final decision is to merge @scottbelden changes?

pkoch:
The decision is yours, but I vote yes.

PR #40, opened 2015-10-12
Support streaming IO by removing need for seek in skip_sync

### LINKED ISSUE #39: Support for reading from streaming IO
At the moment fastavro.reader calls 'seek' in `skip_sync` to seek back 16 bytes if it doesn't find an expected sync marker. This stops it working with streaming IO (e.g. from an HTTP connection), which is a shame as avro is really designed to be streamable.

Is this necessary?

From the spec it looks like the sync marker is just there as a data integrity check, you're already told upfront how many records and how many bytes to expect in the current block. So IIUC you should always find the sync marker at the end of the block -- or at least if you don't then it's exception-worthy, not situation where you should rewind and pretend nothing happened.

Currently you're calling skip_sync at the start of every block; you'll always fail to find a sync marker before the first block, really you should skip it after each block. And if you fail to find it, complain.

Does that sound right?

[EDIT: turns out the purpose of the sync marker is to more to allow hadoop to find the nearest block start in a split chunk of an avro file: https://cwiki.apache.org/confluence/display/AVRO/FAQ#FAQ-Whatisthepurposeofthesyncmarkerintheobjectfileformat  -- not something you need to worry about in a straightforward streaming setup starting from the beginning of the file ]

mjwillson:
This moves skip_sync to the end of the block (where the sync marker exists) rather than the start (where it will exist only when there was a previous block), and has it require the sync marker is present, and not try to handle the case where it isn't present. This means we no longer depend on seek, and can support streaming IO.

Fixes #39 for me.

### DIFF
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -7,7 +7,6 @@
 # Apache 2.0 license (http://www.apache.org/licenses/LICENSE-2.0)
 
 import json
-from os import SEEK_CUR
 from struct import unpack
 from zlib import decompress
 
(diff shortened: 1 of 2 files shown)


tebeka:
Thanks! Will check soon-ish...

lidialiker:
Hi, and thanks a lot for this patch.
It's also fixing the issue when there is a corrupted sync marker at the end of file.
I ran into this recently, in that case fastavro is entering infinite loop because of seeking back.

tebeka:
Thanks. In 0.9.6

PR #41, opened 2015-11-04
Add a fixed-size integer type extension and path-based error handling

dacjames:
This adds two new features that my organization required.
#### Extension for fixed-size integer types

When enabled via the enable_extensions argument to reader/writer, fastavro will now treat fixed types that have names matching the fixed integer types (uint8_t through uint64_t and int8_t through int64_t) as integers. We work with networking devices, where fixed size integers are important.
#### Path-based error reporting

Fastavro now tracks it's position as it traverses a record so that parsing errors provide meaningful error messages. We use fastavro together with a schema registry instead of for reading avro files with an embedded schema, so schema errors are more likely. This feature is critical to help users diagnose where there messages failed to match the expected schema.

Every attempt has been made to follow conventions of other code in the package: tests for these features are provided and flake8 passes. If you wish to pull in one feature and not the other, the commits are relatively isolated; let me know if they need to be isolated further.

### DIFF
--- a/fastavro/errors.py
+++ b/fastavro/errors.py
@@ -1,0 +1,24 @@
+from .util import path_string
+
+
+class UnknownType(Exception):
+    def __init__(self, name):
+        super(UnknownType, self).__init__(name)
+        self.name = name
+
+
+class AvroValueError(ValueError):
+    pass
+
+    @classmethod
+    def create(cls, reading_or_writing, path, traceback):
+        message = '\n'.join([
+            "Error encountered %s avro message." % reading_or_writing,
+            "Path was: %s" % path_string(path),
+            "Exception traceback: %s" % traceback
+        ])
+        return cls(message)
+
+
+class SchemaResolutionError(Exception):
+    pass
--- a/fastavro/extension.py
+++ b/fastavro/extension.py
@@ -1,0 +1,15 @@
+# cython: auto_cpdef=True
+
+'''Avro Extensions'''
+
+FIXED_EXTENSIONS = {
+    # name  : (size, pack/unpack fmt string)
+    'int8_t': (1, 'b'),
+    'int16_t': (2, 'h'),
+    'int32_t': (4, 'i'),
+    'int64_t': (8, 'l'),
+    'uint8_t': (1, 'B'),
+    'uint16_t': (2, 'H'),
+    'uint32_t': (4, 'I'),
+    'uint64_t': (8, 'L'),
+}
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -488,9 +488,9 @@ except ImportError:
 def acquaint_schema(schema,
                     repo=READERS,
                     reader_schema_defs=SCHEMA_DEFS,
-                    enable_enxtensions=False):
+                    enable_extensions=False):
 
-    if enable_enxtensions and 'fixed' in repo:
+    if enable_extensions and 'fixed' in repo:
         repo['fixed'] = read_fixed_extension
 
     extract_named_schemas_into_repo(
@@ -540,7 +540,7 @@ class iter_avro:
             for record in avro:
                 process_record(record)
     '''
-    def __init__(self, fo, reader_schema=None, enable_enxtensions=False):
+    def __init__(self, fo, reader_schema=None, enable_extensions=False):
         self.fo = fo
         try:
             self._header = read_data(fo, HEADER_SCHEMA)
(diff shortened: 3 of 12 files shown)


tebeka:
Thanks! Will look at it soon-ish.

dacjames:
Looks like I have some Python3 problems.  Will get those fixed.

tebeka:
Thanks @dacjames

dacjames:
Fixed all of the Travis issues.

tebeka:
Thanks. Super swamped right now, will try to get to this by end of next week.

fastavro/errors.py
```diff
@@ -0,0 +1,24 @@
...
+class UnknownType(Exception):
+    def __init__(self, name):
+        super(UnknownType, self).__init__(name)
+        self.name = name
+
+
+class AvroValueError(ValueError):
+    pass
+
+    @classmethod
```
tebeka: Why is this a `classmethod` and not a regular `__init__`?
Also class method which are alternate initializers are usually named `from...`

fastavro/errors.py
```diff
@@ -0,0 +1,24 @@
...
+    def create(cls, reading_or_writing, path, traceback):
```
tebeka: `reading_or_writing` -> `action`

fastavro/extension.py
```diff
@@ -0,0 +1,15 @@
+# cython: auto_cpdef=True
+
+'''Avro Extensions'''
+
+FIXED_EXTENSIONS = {
+    # name  : (size, pack/unpack fmt string)
+    'int8_t': (1, 'b'),
```
tebeka: Why the `_t` prefix?

fastavro/reader.py
```diff
@@ -371,13 +427,24 @@ def read_record(fo, writer_schema, reader_schema=None):
 }
 
 
-def read_data(fo, writer_schema, reader_schema=None):
+def read_data(fo, writer_schema, reader_schema=None, path=None):
     '''Read data from file object according to schema.'''
+    if path is None:
```
tebeka:
```python
path = [] if path is None else path
```

fastavro/util.py
```diff
@@ -0,0 +1,72 @@
...
+import sys
+from functools import wraps
+
+if sys.version_info < (3,):
+    INTEGER_TYPES = (int, long,)  # flake8: noqa
+else:
+    INTEGER_TYPES = (int,)
+
+
+def push_path(path, name):
```
tebeka: I don't think this (and `pop_path`) are needed. You can use slicing:
`
```python
path = []  # Empty path
path += [name]  # push
path = path[:-1] # pop
```

fastavro/util.py
```diff
@@ -0,0 +1,72 @@
...
+def pop_path(path):
+    if path is None:
+        return None
+    else:
+        path.pop()
+        return path
+
+
+def path_string(path):
+    def format_item(item):
```
tebeka: This will define a new function every time the function is called and potentially call it. In order maintain speed, I prefer we'll unroll this. Also I `[]` will be the empty path and not `None` it'll make things easier.

```python
fmt = '[%s]' if isinstance(item, INTEGER_TYPE) else '%s'
return '.'.join(fmt % item for item in path)
```

Also what's the motivation for `[1]` display? Maybe we can use `%r` to distinguish between `int` and `str`.

tebeka:
Can you separate this to two pull requests. I'm pretty happy with the fixed sized one but on the fence with the path one.

fastavro/extension.py
dacjames: To match <stdint.h>, which defines fixed size types in C.

fastavro/util.py
dacjames: I fielded both options and my team just preferred `[0]` for indexing.

fastavro/reader.py
dacjames: Why? Same behavior, but less obvious.

fastavro/errors.py
dacjames: To maintain the same interface as normal exceptions: `AvroValueError("some message")` should behave like `TypeError("some message")`, `ValueError("some message")`, etc. I quite dislike when libraries break this expectation.

Agree on the name, though.

fastavro/util.py
dacjames: Regarding the need for push_path and family: setting default parameters to mutable objects like list   leads to all kinds of unexpected behavior. Making the path a required parameter would fix this, though, but it changes the interface to all the read_* and write_* functions, which I didn't want to do.

dacjames:
I'm going to be on vacation for a few weeks, but I'll split up the commits.

I'm open to other suggests on how to implement the error handling. I don't particularly like the current implementation (and the overhead it might introduce), but I couldn't think of anything better. This is a critical feature for when you're using Avro as an network format rather than a file format because users mess up the message format and the random exceptions thrown currently don't provide any clue where you made the mistake.

tebeka:
Thanks! I'll think of how to improve error handling.

Have fun in your vacation.

PR #42, opened 2015-12-08
adding support for python 2.6

antonymayi:
python 2.6 is still heavily used thanks to redhat/centos distributions so supporting it is useful.

this PR makes two minor changes to add the support:
* replacing relative imports


antonymayi:
will resubmit

PR #43, opened 2015-12-08
python 2.6 compatibility

antonymayi:
python 2.6 is still used a lot thanks to RHEL/Centos so nice to support it.

making two minor changes:
* replacing relative imports


antonymayi:
will resubmit

PR #44, opened 2015-12-08
python 2.6 compatibility

antonymayi:
python 2.6 is still used a lot thanks to RHEL/Centos distros so supporting it is useful.

making two minor changes to add the support:
* replacing relative imports


tebeka:
Thanks will take a look soon-ish.

Sorry for the late reply, just came back from vacation.

Issue #49, opened 2016-05-19
fastavro gives a default value for unions with nulls that don't have a default value


movermeyer:
This is somewhat related to #48. Given this example schema:

```
protocol test {
  record a {
    int x;
    union {null, int} y;
  }
}
```

the following should fail:

```
>>> data = StringIO.StringIO()
>>> fastavro.writer(data, schema, [{'val': {'x': 1}}])
```

However, it does not, giving a default value of null to a field ('y') without a default value.

This stems from [the use of .get()](https://github.com/tebeka/fastavro/blob/53648c81598f40765ef6bd48923a92255ebcc429/fastavro/writer.py#L241), which returns None if the key is not found. 
The field does not have a default value, so datum.get('default') returns None, which is then used as the default value.

This is just mixing up the lack of a value, with a value of null. Instead, the code should check for the field's existence in the datum, and return false if it doesn't exist.

This is related to #37, in that having write_record call validate() on itself would catch this. Although I imagine the overhead might raise some ire, so there are some options.

Some pull requests will be made in a bit.

movermeyer:
In general, this is a problem with not checking to make sure that the datum contains a value for each field or the field has a default.

In this specific case, it results in data being made up.

movermeyer:
I like @e-heller's solution better. Solves this problem and shouldn't have noticeable impact on performance.

tebeka:
0.9.10 should fix this

movermeyer:
Looks good. Thanks.

PR #50, opened 2016-05-19
A potential fix for issue #49.

movermeyer:
Have the record validate itself before writing out.

We don't technically need to do this level of validation. It was done this way to avoid duplicating code between validate([#L205](https://github.com/tebeka/fastavro/blob/53648c81598f40765ef6bd48923a92255ebcc429/fastavro/writer.py#L205)) and write_record ([#L241](https://github.com/tebeka/fastavro/blob/53648c81598f40765ef6bd48923a92255ebcc429/fastavro/writer.py#L241)), as both need to do the check. 

I think the only case where this is a problem is the case where the field is a union with null. So there is probably a way to rework this to have less overhead. But I'll leave this to your direction @tebeka.

### DIFF
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -202,8 +202,8 @@ def validate(datum, schema):
         return (
             isinstance(datum, Mapping) and
             all(
-                validate(datum.get(f['name'], f.get('default')), f['type']) 
-                if not (f['name'] not in datum and 
+                validate(datum.get(f['name'], f.get('default')), f['type'])
+                if not (f['name'] not in datum and
                         'default' not in f) else False
                 for f in schema['fields']
             )
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -142,6 +142,8 @@ INT_MAX_VALUE = (1 << 31) - 1
 LONG_MIN_VALUE = -(1 << 63)
 LONG_MAX_VALUE = (1 << 63) - 1
 
+def no_value_and_no_default(datum, field):
+    return field['name'] not in datum and 'default' not in field
 
 def validate(datum, schema):
     """Determine if a python datum is an instance of a schema."""
@@ -202,9 +204,7 @@ def validate(datum, schema):
         return (
             isinstance(datum, Mapping) and
             all(
-                validate(datum.get(f['name'], f.get('default')), f['type'])
-                if not (f['name'] not in datum and
-                        'default' not in f) else False
+                (not no_value_and_no_default(datum, f)) and validate(datum.get(f['name'], f.get('default')), f['type'])
                 for f in schema['fields']
             )
         )
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -202,7 +202,7 @@ def validate(datum, schema):
         return (
             isinstance(datum, Mapping) and
             all(
-                validate(datum.get(f['name']), f['type'])
+                validate(datum.get(f['name'], f.get('default')), f['type']) if not (f['name'] not in datum and 'default' not in f) else False
                 for f in schema['fields']
             )
         )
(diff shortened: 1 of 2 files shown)


e-heller:
It might be simpler to define a sentinel object for when there is "no value"

e.g., define the sentinel `NoValue`:
```python
NoValue = object()
```

Then in the `validate` function you can do:
```python
if record_type in ('record', 'error', 'request',):
    return (
        isinstance(datum, Mapping) and
        all(
            validate(datum.get(f['name'], f.get('default', NoValue)),
                     f['type'])
            for f in schema['fields']
        )
    )
```
This works because `NoValue` will not successfully validate against any schema type, so an exception will be raised.

In `write_record` there will be no need to call `validate()` -- using `NoValue` will be sufficient to ensure an exception is raised:
```python
def write_record(fo, datum, schema):
    for field in schema['fields']:
        write_data(fo,
                   datum.get(field['name'], field.get('default', NoValue)),
                   field['type'])
```

fastavro/writer.py
```diff
@@ -202,7 +202,9 @@ def validate(datum, schema):
         return (
             isinstance(datum, Mapping) and
             all(
-                validate(datum.get(f['name']), f['type'])
+                validate(datum.get(f['name'], f.get('default')), f['type'])
+                if not (f['name'] not in datum and
```
tebeka: * I find this logic hard to read. Is there a way to create a utility function that will make this code clearer?
* Not sure if we want to fail. There [is a definition for type defaults](https://avro.apache.org/docs/1.8.0/spec.html#schema_record) even when there is no default value.

fastavro/writer.py
```diff
@@ -236,6 +238,11 @@ def write_record(fo, datum, schema):
     that they are declared. In other words, a record is encoded as just the
     concatenation of the encodings of its fields.  Field values are encoded per
     their schema."""
+
+    if not validate(datum, schema):
```
tebeka: In the past I didn't want to do this due to speed concerns. You can validate yourself before writing.

I *might* add an optional value `validate` which if set (default to `False`) will validate.

movermeyer:
I prefer @e-heller's [solution](https://github.com/e-heller/fastavro/commit/80282e0abb0df1ab0345f38717dd2932268efa19)

tebeka:
@movermeyer Can you fix the tests (see [here](https://travis-ci.org/tebeka/fastavro/jobs/132543649))

movermeyer:
Done.

scottbelden:
I'm going to close this as it conflicts with the previously merged code [here](https://github.com/tebeka/fastavro/pull/70).

It's hard to say which approach is _right_. Should we for people to explicity create a record like `{'foo': None}` for a schema where `foo` is `['null', 'int']` or since it can be null do we allow a no-value to mean the same as it being `None`? Arguments could be made both ways, but it should be noted that the current approach (i.e. allow no-value to be `None`) is the approach that is taken by the standard python avro library:

```python
In [1]: from io import BytesIO
   ...: import avro.schema
   ...: from avro.datafile import DataFileReader, DataFileWriter
   ...: from avro.io import DatumReader, DatumWriter
   ...: 
   ...: schema = {
   ...:     "type": "record",
   ...:     "name": "foo",
   ...:     "fields": [{
   ...:         "name": "x",
   ...:         "type": ["int", "null"]
   ...:     }]
   ...: }
   ...: 
   ...: iostream = BytesIO()
   ...: writer = DataFileWriter(iostream, DatumWriter(), avro.schema.SchemaFromJSONData(schema))
   ...: writer.append({})
   ...: writer.flush()
   ...: 
   ...: iostream.seek(0)
   ...: records = list(DataFileReader(iostream, DatumReader()))
   ...: print(records)
   ...: 
[{'x': None}]
```

PR #52, opened 2016-05-27
better stacktrace info on write errors

scottbelden:
This is a subset of https://github.com/tebeka/fastavro/pull/41 that only includes the error path tracking for write operations. I unrolled some of the decorators from that PR (which while much more elegant, probably added more overhead).

An example stacktrace looks like this:
```python
  File "/Users/scott/.virtualenvs/fastavro/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/scott/git/scottbelden/fastavro/tests/test_fastavro.py", line 329, in test_repo_caching_issue
    fastavro.writer(new_file, other_schema, records)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 457, in writer
    write_data(io, record, schema, path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 310, in write_data
    path=path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 254, in write_record
    path=path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 310, in write_data
    path=path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 375, in <lambda>
    path),
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 310, in write_data
    path=path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 254, in write_record
    path=path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 310, in write_data
    path=path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 254, in write_record
    path=path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 315, in write_data
    raise AvroValueError.from_traceback(datum, schema, path, tb)
AvroValueError: Exception traceback: Traceback (most recent call last):
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 310, in write_data
    path=path)
  File "/Users/scott/git/scottbelden/fastavro/fastavro/writer.py", line 51, in write_int
    datum = (datum << 1) ^ (datum >> 63)
TypeError: unsupported operand type(s) for <<: 'str' and 'int'

Path is: <record[0]>.<field[aa]>.<field[b]>.<field[c]>
Datum is 2 of type <type 'str'>
Expected avro schema type: int
```

I did some profiling and the results are here: https://gist.github.com/scottbelden/418e2523dee10da0798978ec97226627

I would ignore the overall timing and rather just note that every function call has the exact same number of calls except of course ```{method 'append' of 'list' objects}``` which is only present in the run which included the changes in this PR. While there are 336,365 calls, it doesn't add that much overhead (i.e. 32 thousandths of a second).

Please let me know what you think.

### DIFF
--- a/fastavro/exceptions.py
+++ b/fastavro/exceptions.py
@@ -1,0 +1,10 @@
+class AvroValueError(ValueError):
+    @classmethod
+    def from_traceback(cls, datum, schema, path, traceback):
+        msg = '\n'.join([
+            'Exception traceback: %s' % traceback,
+            'Path is: %s' % '.'.join(path),
+            'Datum is %s of type %s' % (datum, type(datum)),
+            'Expected avro schema type: %s' % schema,
+        ])
+        return cls(msg)
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -7,11 +7,13 @@
 # Apache 2.0 license (http://www.apache.org/licenses/LICENSE-2.0)
 
 try:
+    from fastavro._exceptions import AvroValueError
     from fastavro._six import utob, MemoryIO, long, is_str, iteritems
     from fastavro._reader import HEADER_SCHEMA, SYNC_SIZE, MAGIC
     from fastavro._schema import extract_named_schemas_into_repo,\
         extract_record_type
 except ImportError:
+    from fastavro.exceptions import AvroValueError
     from fastavro.six import utob, MemoryIO, long, is_str, iteritems
     from fastavro.reader import HEADER_SCHEMA, SYNC_SIZE, MAGIC
     from fastavro.schema import extract_named_schemas_into_repo,\
@@ -26,23 +28,24 @@ from binascii import crc32
 from collections import Iterable, Mapping
 from os import urandom, SEEK_SET
 from struct import pack
+import traceback
 from zlib import compress
 
 NoneType = type(None)
 
 
-def write_null(fo, datum, schema=None):
+def write_null(fo, datum, **kwargs):
     """null is written as zero bytes"""
     pass
 
 
-def write_boolean(fo, datum, schema=None):
+def write_boolean(fo, datum, **kwargs):
     """A boolean is written as a single byte whose value is either 0 (false) or
     1 (true)."""
     fo.write(pack('B', 1 if datum else 0))
 
 
-def write_int(fo, datum, schema=None):
+def write_int(fo, datum, **kwargs):
     """int and long values are written using variable-length, zig-zag coding.
     """
     datum = (datum << 1) ^ (datum >> 63)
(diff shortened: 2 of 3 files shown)


scottbelden:
After thinking about this some more, I think there might be a better way to go about this. Instead of keeping track of the path information as the serialization happens, it would probably be better to have wrap the outer most calls in a try/except and if it hits an exception then execute a validation function (more thorough than just the ```writer.validate()```) that will walk through the schema and figure out the problem.

This is a bit redundant since it would walk the data structure and schema twice, but I don't think there is a need to be as efficient when an exception happens. Therefore, the actual serialization code can stay fast and if an exception happens, the less efficient but thorough data validation can take place and pin point exactly where the problem is.

I should be able to make a PR this weekend that can show that approach.

scottbelden:
Alternative approach here: https://github.com/tebeka/fastavro/pull/53

andrusha:
Error messages still aren't all that useful, it is impossible to determine which part of the schema triggered the error. E.g.:

```
  File "fastavro/_writer.py", line 439, in fastavro._writer.writer (fastavro/_writer.c:9956)
  File "fastavro/_writer.py", line 291, in fastavro._writer.write_data (fastavro/_writer.c:7311)
  File "fastavro/_writer.py", line 234, in fastavro._writer.write_record (fastavro/_writer.c:7203)
  File "fastavro/_writer.py", line 243, in fastavro._writer.write_record (fastavro/_writer.c:7079)
  File "fastavro/_writer.py", line 291, in fastavro._writer.write_data (fastavro/_writer.c:7311)
  File "fastavro/_writer.py", line 216, in fastavro._writer.write_union (fastavro/_writer.c:6884)
  File "fastavro/_writer.py", line 227, in fastavro._writer.write_union (fastavro/_writer.c:6708)
ValueError: ['b', 'b'] (type <class 'list'>) do not match ['null', 'string']
```

scottbelden:
That's probably because it's just using the normal validate function which doesn't include that bit of useful information. You could steal this validate function (https://github.com/scottbelden/fastavro/blob/415a37a07b8bfe71bef852da494eae9c0fee3c1a/fastavro/validate.py#L29-L152) which is from https://github.com/tebeka/fastavro/pull/53 and it would probably give you the type of validation you are looking for.

However, that validate function is expensive and not really ideal to run before every write call. My suggestion would be to wrap the fastavro.write call in a try/except and run the better validator on exception. Technically it would add a tiny bit of overhead, but I feel like that overhead is worth it and is what I do in my fork of fastavro.

PR #53, opened 2016-05-30
better error information when serialization fails

scottbelden:
Alternative approach to https://github.com/tebeka/fastavro/pull/52 that doesn't add extra overhead to the serialization and instead re-walks the schema and datum to determine the problem when serialization fails.

There are still a few pieces that could be polished a little like the printing of union errors, but just putting this up here to let travis test it and get feedback on this approach vs https://github.com/tebeka/fastavro/pull/52.

### DIFF
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -7,15 +7,21 @@
 # Apache 2.0 license (http://www.apache.org/licenses/LICENSE-2.0)
 
 try:
+    from fastavro._constants import INT_MIN_VALUE, INT_MAX_VALUE,\
+        LONG_MIN_VALUE, LONG_MAX_VALUE
     from fastavro._six import utob, MemoryIO, long, is_str, iteritems
     from fastavro._reader import HEADER_SCHEMA, SYNC_SIZE, MAGIC
     from fastavro._schema import extract_named_schemas_into_repo,\
         extract_record_type
+    from fastavro._validate import validate_data
 except ImportError:
+    from fastavro.constants import INT_MIN_VALUE, INT_MAX_VALUE,\
+        LONG_MIN_VALUE, LONG_MAX_VALUE
     from fastavro.six import utob, MemoryIO, long, is_str, iteritems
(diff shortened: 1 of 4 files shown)


fastavro/writer.py
```diff
@@ -425,7 +425,11 @@ def dump():
     acquaint_schema(schema)
 
     for record in records:
-        write_data(io, record, schema)
+        try:
```
tebeka: I need to think about this one. Looks like a good start, however:
* The name `validate_data` does not really fit what the function is doing.
* The error raised by `validate_data` might mask the original error (`Exception`)
* `try/catch` has performance penalty, we need to benchmark

scottbelden: You are correct that the original exception would get swallowed. This was done intentionally since I thought the exception from ```validate_data``` would be superior, but thinking about it again I think it's good to keep the original exception, so that can be fixed.

As for the benchmarking, I'll try to do some and post the results. From what I've read, try/excepts that don't raise an exception are extremely cheap, but I'll see if I can get some measurements.

scottbelden: It looks like the try/except will add about .015 usec of overhead (per record)

```
python -m timeit 'try:' '  str(1)' 'except Exception:' '  pass'
10000000 loops, best of 3: 0.131 usec per loop

python -m timeit 'str(1)'
10000000 loops, best of 3: 0.115 usec per loop
```

tebeka: My current idea is to add an optional for validation.

```
In [1]: %timeit pass  
100000000 loops, best of 3: 9.64 ns per loop

In [2]: val = True  

In [3]: %timeit if val: pass  
10000000 loops, best of 3: 27.1 ns per loop
```

scottbelden: Where would you add the ```if val:```? I don't follow.

tebeka:
Added `validator` optional argument to `fastavro.writer.write`. Will come out in 0.9.11 shortly.

PR #56, opened 2016-07-01
Load schema recursively

### LINKED ISSUE #55: Composable schemas
The Java Avro Maven plugin lets you compose schemas that may be defined across multiple files using an `<import>` feature. I was able to do something similar using fastavro with a little utility code:

```
def load_schema_recursively(schema_file, schema=None):                                 
    if schema is None:
        with open('schema_files/src/main/avro/{}.avsc'.format(topic)) as fd:           
            schema = json.load(fd)
    try:  
        fastavro.acquaint_schema(schema)
    except UnknownType as e:
        child_name = e.name.split('.')[1]
        load_schema_recursively(child_name)
        load_schema_recursively(schema_file, schema)
    return schema
```
Would this kind of functionality be useful in fastavro or is this outside of the scope of what this lib does?

natb1:
resolves #55 

The way it checks to see if a schema is already registered is probably an abuse of the UnkownType exception. Probably not a big deal but let me know if you would like to do that check another way.

### DIFF
--- a/fastavro/schema.py
+++ b/fastavro/schema.py
@@ -121,6 +121,9 @@ def _load_schema(schema, schema_dir):
     try:
         acquaint_schema(schema)  # NOQA
     except UnknownType as e:
-        load_schema(os.path.join(schema_dir, '{}.avsc'.format(e.name)))
+        try:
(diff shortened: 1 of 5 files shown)


tebeka:
Thanks!

PR #59, opened 2016-07-22
Made fix to reader.py for reader schemas with default values of None.…

kurtostfeld:
… Added corresponding unit tests

### DIFF
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -341,7 +341,7 @@ def read_record(fo, writer_schema, reader_schema=None):
             for field_name, field in iteritems(readers_field_dict):
                 if field_name not in writer_fields:
                     default = field.get('default')
-                    if default:
+                    if 'default' in field:
                         record[field['name']] = default
                     else:
                         msg = 'No default value for %s' % field['name']
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -491,8 +491,7 @@ class iter_avro:
     def next(self):
         return next(self._records)
 
-    def __next__(self):
-        return next(self._records)
+    __next__ = next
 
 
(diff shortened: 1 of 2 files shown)


scottbelden:
Nice catch.

tebeka:
Thanks! Will have a look soon-ish.

tests/test_schema_evolution.py
```diff
@@ -48,7 +48,7 @@ def avro_to_bytes_with_schema(avro_schema, avro_dict):
 def bytes_with_schema_to_avro(avro_read_schema, binary):
     with BytesIO(binary) as bytes_io:
         reader = fastavro.reader(bytes_io, avro_read_schema)
-        return next(reader)
+        return reader.next()
```
tebeka: I don't understand how this fails. In Python 3 we use `obj.__next__()` while in Python 2 it's `obj.next()`. The builtin `next` should talk care of this.

@kurtostfeld Can you shed some light on this?

kurtostfeld: The iter_avro class in the master branch on this git repo does not define a __next__() function. It just defines Python 2 style next()

tebeka: Good catch! Then this should be the fix - do you want to make it or should I?

fastavro/reader.py
```diff
@@ -491,6 +491,9 @@ def __iter__(self):
     def next(self):
         return next(self._records)
 
+    def __next__(self):
```
tebeka: I prefer

```python
    __next__ = next
```

kurtostfeld: I agree. updated.

kurtostfeld:
awesome! thank you guys :)

PR #61, opened 2016-09-28
Refactor writer as optional class

qix:
I'm using fastavro to write out large files (hundreds of mb+)

At that point the speed savings start becoming significant, but also I don't want to store the entire file in memory. Unfortunately the `writer()` api requires a list of records.

I refactored it to a class, with minimal changes. Unfortunately it requires that you call `.flush()` which is not an ideal api.

### DIFF
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -354,6 +354,55 @@ def acquaint_schema(schema, repo=None):
     )
 
 
+class Writer(object):
+    def __init__(self,
+                 fo,
+                 schema,
+                 codec='null',
+                 sync_interval=1000 * SYNC_SIZE,
+                 metadata=None,
+                 validator=None):
+        self.fo = fo
+        self.schema = schema
+        self.validate_fn = validate if validator is True else validator
+        self.sync_marker = urandom(SYNC_SIZE)
+        self.io = MemoryIO()
+        self.block_count = 0
+        self.metadata = metadata or {}
+        self.metadata['avro.codec'] = codec
+        self.metadata['avro.schema'] = json.dumps(schema)
+        self.sync_interval = sync_interval
+
+        try:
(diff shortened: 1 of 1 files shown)


tebeka:
Thanks @qix . However the `records` parameter to writer is an iterable. It doesn't have to be a list but can be a generator that produces records one at a time. Unless I'm missing something?

qix:
Sure that would work as long writing a single avro file is the primary purpose of your application.

Writing any avro on the side, or for our case trying to write multiple files in parallel, is not supported.

qix:
We're looking at open sourcing the project, but for now here is a dump of the code if you're interested:

https://github.com/qix/kafka-store-tmp/blob/master/kafka_store/handler.py

Essentially it reads from kafka on multiple partitions, and dumps each partition to an avro file in an efficient and atomic (only once) manor.

tebeka:
By "parallel" do you mean multi-threading or several files each from a source?

I see in the example that you write records as they come. This can be done with a writer thread, something like the (untested) code below:

```python
from threading import Thread
from queue import Queue

queue = Queue()
Thread(target=fastavro.writer, args=(fo, schema, iter(queue.get, None), daemon=True).start()
```
Then push records to queue when they arrive. Signal done with pushing None to the queue.

Having said all that. I'm OK with merging your code - just make it pass the test suite first please :)

qix:
I fought with the test suite a little bit. It looks like the problem is `flake8` dropped support for `python2.6`. I tried to force it to use the last supported version, but I'll open a seperate pull request for that.

You might just want to remove python2.6 from supported versions, but leave that up to you.

qix:
(i.e. my change did not break them, they're currently failing in master.)

It'd be great if I could get this merged upstream.

PR #64, opened 2016-11-24
Fix block count corruption on writer.flush() after an auto-sync

DouglasOrr:
Small change - I think it seems right to reset the block count to zero whenever you do `writer.dump()` (whether through an auto-sync or a manual `flush()`). Otherwise `dump()` can be called with no available blocks, which seems to be bad.

### DIFF
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -387,6 +387,7 @@ class Writer(object):
         self.fo.write(self.sync_marker)
         self.io.truncate(0)
         self.io.seek(0, SEEK_SET)
+        self.block_count = 0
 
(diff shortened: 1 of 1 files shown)


tebeka:
Thanks! 

For some seemingly unrelated problem travis is failing. I'll run some tests locally and fix this before final merge.

tebeka:
Can you merge with master so I'll see travis passing? Thanks

tebeka:
Thanks!

DouglasOrr:
Cheers (and sorry I didn't get a chance to do the merge for Travis)!

PR #65, opened 2016-11-25
Don't make PyPy guys sad :(

TiberiumPY:
Why do you correctly spell Python with first capital letter, but not PyPy?


tebeka:
I spelled like the logo is spelling. Fixed - thanks.

PR #70, opened 2017-04-16
write None value for nullable field

nobo728x:
ValueError is raised for nullable field

### DIFF
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -146,7 +146,6 @@ LONG_MAX_VALUE = (1 << 63) - 1
 
 def validate(datum, schema):
     """Determine if a python datum is an instance of a schema."""
-
     record_type = extract_record_type(schema)
 
     if record_type == 'null':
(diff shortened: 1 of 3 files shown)


tebeka:
Thanks for improving fastavro!
I'm on the road currently, will look at it end next week.

fastavro/writer.py
```diff
@@ -239,9 +238,15 @@ def write_record(fo, datum, schema):
     their schema."""
     for field in schema['fields']:
         name = field['name']
+        f_type = field['type']
         if name not in datum and 'default' not in field:
-            raise ValueError('no value and no default for %s' % name)
-        write_data(fo, datum.get(name, field.get('default')), field['type'])
+            if 'null' in f_type:
```
scottbelden: I think you can accomplish what you are looking for by just changing the original 
```python
if name not in datum and 'default' not in field:
```

to

```python
if name not in datum and 'default' not in field and 'null' not in field['type']:
```

nobo728x: You are right :+1:

nobo728x: Change done.

Issue #58, opened 2016-07-22
Performance Optimizations: Function calls are taxing


brianbruggeman:
Python incurs significant overhead when functions are called.  

Looking at just the writer functionality...

One clear optimization here is not to create functions (e.g. Null/None) when they do nothing.  Another option is to possibly replace functions that are just wrappers (.e.g utf8, long and bytes).  Probably the most performant optimization is to simply use an if block rather than a map with a function call.  You might modify the mapping to identify the struct pack character rather than calling a function.

Based on my quick look at a `cython -a <filename>`, and then opening up the html file, I can see that all of the functions are still using their python form rather than a pure C form.  So one of the worst performance problems of python (function calls) is not being replaced by Cython.  That tells me there's a lot of room for improvement.

The Cython solution is to create a hybrid cython file:  pyx.  And to really focus on creating types and memory allocation.  With enough massaging, it won't look much like python at all (nor will it look like C) but it will be fast.  It also ends up ruining DRY and you'll have to maintain two files.

However, if you have to maintain two files anyway, the better approach is probably to move to C or possibly Python's C-API for the slow pieces.

tebeka:
Thanks for reporting! This is a big task which current I don't have the bandwidth to do. Switching from Cython to C is something I need to think about. There are many other consideration other than speed, and C makes it easy to make a lot of mistakes.

Another option I was considering was wrapping the [Avro C library](https://avro.apache.org/docs/current/api/c/) with Cython.

As always - patches are welcome :)

brianbruggeman:
I looked (a little) into wrapping the Avro C Library.  I think that it's a viable approach.

In wrapping, you still have multiple choices.

libffi -- you get to retain compatibility with pypy at the expense of a long warmup time for cpython and overhead while executing
ctypes -- is more performant for cpython but still has overhead
python c-api -- no compatibility with pypy, but highly performant

For what it's worth, I'm currently looking at a 10 ms window for one of my data sources and in that window, I need to receive the data on a REST API, store it to a local file and then create a response.  I've created a lua plugin for nginx to stay within that window.  But, I'd also like to have a side process which reads my local file and sends it to kafka with the avro encoding.  I'm not necessarily constrained by that 10 ms window, but I do not want to overload the system with avro encoding... so speed is still important.

[...]

Issue #51, opened 2016-05-25
High CPU Utilisation


mohitapplift:
I am using fastavro for processing live log records. The frequency at which I am ingesting log records is around 1000 records/second. The problem I am facing is high CPU utilization. CPU utilisation is always ranging around 100%. Any suggestion is most welcome

tebeka:
@mohitapplift Can you provide a test case?

barrywhart:
This doesn't seem like a problem per se.  A heavily loaded application will generally max out *some* resource, whether CPU, disk, network, etc.

It sounds like you wish the library was faster.

In the absence of a faster implementation, you might be able to parallelize the processing (e.g. two subprocesses, each processing 500 records/second).

scottbelden:
Agree with @barrywhart. I'm going to close this. If you believe there are bugs that are causing unnecessary computation please create a new issue.

PR #77, opened 2017-05-17
Adds record writer for tuples and lists

### LINKED ISSUE #76: Allow writer to support lists or tuples (or named tuples) in addition to dict as record type
The [writer](https://github.com/tebeka/fastavro/blob/8d304a6084130fd152fd892153327df5e70363b8/fastavro/writer.py#L428-L483) currently requires a dictionary to process each record entry, but the documentation indicates that it's okay to use an iterable, with no requirement imposed on data structure.

```
In [1]: import fastavro

In [2]: schema = {"type": "record", "fields": [{"name": "name", "type": "string"}, {"name": "type", "type": "string"}]}

In [3]: records = [
   ...:     ("apple", "fruit"),
   ...:     ("banana", "fruit"),
   ...: ]

In [4]: path = "tmp.avro"

In [5]: with open(path, 'wb') as stream:
   ...:     fastavro.writer(stream, schema, records)
   ...:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-3c54c57bc598> in <module>()
      1 with open(path, 'wb') as stream:
----> 2     fastavro.writer(stream, schema, records)
      3

<redacted>/site-packages/fastavro/_writer.so in fastavro._writer.writer (fastavro/_writer.c:12273)()

<redacted>/site-packages/fastavro/_writer.so in fastavro._writer.writer (fastavro/_writer.c:12052)()

<redacted>/site-packages/fastavro/_writer.so in fastavro._writer.Writer.write (fastavro/_writer.c:11527)()

<redacted>/site-packages/fastavro/_writer.so in fastavro._writer.write_data (fastavro/_writer.c:8450)()

<redacted>/site-packages/fastavro/_writer.so in fastavro._writer.write_record (fastavro/_writer.c:8310)()

<redacted>/site-packages/fastavro/_writer.so in fastavro._writer.write_record (fastavro/_writer.c:8117)()

ValueError: no value and no default for name

In [6]: records_as_dict = [
   ...:     {'name': 'apple', 'type': 'fruit'},
   ...:     {'name': 'banana', 'type': 'fruit'},
   ...: ]

In [7]: with open(path, 'wb') as stream:
   ...:     fastavro.writer(stream, schema, records_as_dict)
   ...:
```

It would make sense that a tuple, list or namedtuple would be an easy (and possibly preferred) data structure to be used in addition to a dictionary.  This is especially true if the data is already in a column-based format (e.g.  `string.split('\t')`, csv read, db read, etc.).

brianbruggeman:
Fixes #76 

It appears that timing wise, the lower and upper bound for a percentage hit in handling both dictionaries and lists/tuples is anywhere between 5% and 25%.  Pypy is actually faster...  But I am guessing that the timing there has something to do with how pypy warms up.  I added a benchmarking test, but it doesn't appear that the test seems to work well with being consistent... so I've disabled the checks.

### DIFF
--- a/fastavro/writer.py
+++ b/fastavro/writer.py
@@ -28,6 +28,7 @@ from os import urandom, SEEK_SET
 from struct import pack
 from zlib import compress
 
+
 NoneType = type(None)
 
 
@@ -255,13 +256,30 @@ def write_record(fo, datum, schema):
     that they are declared. In other words, a record is encoded as just the
     concatenation of the encodings of its fields.  Field values are encoded per
     their schema."""
-    for field in schema['fields']:
-        name = field['name']
-        if name not in datum and 'default' not in field and\
-                'null' not in field['type']:
-            raise ValueError('no value and no default for %s' % name)
-        write_data(fo, datum.get(
-            name, field.get('default')), field['type'])
+    fields = schema['fields']
+    # the if statement block below increases the timing anywhere between
+    # 5% and 16%
+    if isinstance(datum, dict):
+        for field in fields:
+            name = field['name']
+            default_value = field.get('default')
+            default_value_is_available = 'default' in field
+            field_is_nullable = 'null' in field['type']
+            if ((name not in datum)
+                    and (not default_value_is_available)
+                    and (not field_is_nullable)):
+                raise ValueError('no value and no default for %s' % name)
+            value = datum.get(name, default_value)
+            write_data(fo, value, field['type'])
+
+    elif isinstance(datum, (list, tuple)):
+        index = 0
+        if not len(datum) == len(fields):
+            err = 'datum mismatch with schema:\n    %s\n    %s'
+            raise ValueError(err % (datum, schema))
+        for index, field in enumerate(fields):
+            value = datum[index]
+            write_data(fo, value, field['type'])
 
 
 WRITERS = {
--- a/tests/test_record_writer.py
+++ b/tests/test_record_writer.py
@@ -1,0 +1,149 @@
+# -*- coding: utf-8 -*-
+from __future__ import (
+    absolute_import,
+    division,
+    print_function,
+    unicode_literals
+)
+
+import os
+import sys
+import tempfile
+from collections import namedtuple
+
+
+temp_folder = tempfile.mkdtemp()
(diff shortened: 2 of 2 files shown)


brianbruggeman:
@tebeka This is probably ready for a merge if you acquiesce to the performance hit.

scottbelden:
I'm not a maintainer and so I don't have much say, but I do use this library in my codebase. This feels to me like it would be better in a fork of the project. This is a performance hit for every user of the library (except apparently the pypy folks) for a use case that 99% don't have.

brianbruggeman:
> This is a performance hit for every user of the library (except apparently the pypy folks) for a use case that 99% don't have.

I'd better believe this if you gave some solid evidence for your numbers here.

tebeka:
Let's try and be civil :) 25% hit seems high, the overhead for converting a tuple to a dict doesn't see that high for me:
```
In [1]: values = (1, 2, 3, 4)

In [2]: columns = ('a', 'b', 'c', 'd')

In [3]: %timeit dict(zip(columns, values))
760 ns ± 6.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
And there's something to be said about keeping the code simpler. I'm currently leaning against this but will try to think a way to speed this up.

brianbruggeman:
Okay.  

I'm done attempting to contribute.  I have already wasted too much time here.  Thank you for making this easy. 

1.  There's no point in getting irritated over a project that says on the one hand we need updates, but then is unwilling to accept performance improvements elsewhere (#6 #51 #58) .  
2.  Is additionally not interested in really pushing anything forward by putting the project on maintenance. 
3.  Maintainer has no real skin in development here anymore
4.  And I have alternatives:  https://pypi.python.org/pypi/quickavro  <-- this is probably where fastavro should be going anyway.

tebeka:
@brianbruggeman Sorry you feel this way. I have very little time to maintain `fastavro` (seeking a new maintainer as the README states) and it's been about 5 years since I've used Avro at work. Having said that - I'm trying to keep it fast and relevant as much as I appreciate any help, sometimes it doesn't align with how I see things. There are many excellent Python bindings to Avro out there, I hope one of them will be better for you.

PR #79, opened 2017-05-18
avro 1.8 logical types support

kkirsanov:
date(int), timestamp-millis(long), timestamp-micros(long), decimal(bytes)

### DIFF
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -9,17 +9,22 @@
 import json
 from struct import unpack, error as StructError
 from zlib import decompress
-
+from os import SEEK_SET
 import datetime
+from decimal import getcontext, Decimal
+
+import struct
 
 try:
-    from fastavro._six import MemoryIO, xrange, btou, utob, iteritems, is_str
+    from fastavro._six import MemoryIO, xrange, btou, utob, iteritems,\
+        is_str, PY3
     from fastavro._schema import (
         extract_record_type, acquaint_schema, populate_schema_defs,
         extract_logical_type
     )
 except ImportError:
-    from fastavro.six import MemoryIO, xrange, btou, utob, iteritems, is_str
+    from fastavro.six import MemoryIO, xrange, btou, utob, iteritems, \
+        is_str, PY3
     from fastavro.schema import (
         extract_record_type, acquaint_schema, populate_schema_defs,
         extract_logical_type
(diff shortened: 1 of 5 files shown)


fastavro/reader.py
```diff
@@ -134,6 +142,87 @@ def read_boolean(fo, writer_schema=None, reader_schema=None):
     return unpack('B', fo.read(1))[0] != 0
 
 
+def read_timestamp_millis(data, writer_schema=None, reader_schema=None):
```
tebeka: Combine this one with `read_timestamp_micros`, add a parameter for the resolution.

fastavro/reader.py
```diff
@@ -134,6 +142,87 @@ def read_boolean(fo, writer_schema=None, reader_schema=None):
...
+    return datetime.date.fromordinal(data)
+
+
+def read_bytes_decimal(data, writer_schema=None, reader_schema=None):
+    scale = writer_schema['scale']
+    precision = writer_schema['precision']
+
+    size = len(data)
+
+    if PY3:
```
tebeka: Instead of having `if` on every function call, do this at definition time (outside of this function)

```python
if PY3:
    def read_decimal_from_fixed(...
else:
    def read_decimal_from_fixed(...
```

fastavro/reader.py
```diff
@@ -9,16 +9,24 @@
...
+import datetime
+from decimal import getcontext, Decimal
+
+import struct
 
 try:
-    from fastavro._six import MemoryIO, xrange, btou, utob, iteritems, is_str
-    from fastavro._schema import (
-        extract_record_type, acquaint_schema, populate_schema_defs
+    from ._six import MemoryIO, xrange, btou, utob, iteritems,\
```
tebeka: Will these relative imports work in Python 2.6? Can you test please?

fastavro/reader.py
```diff
@@ -134,6 +142,87 @@ def read_boolean(fo, writer_schema=None, reader_schema=None):
...
+
+    size = len(data)
+
+    if PY3:
+        unscaled_datum = read_decimal_from_fixed(data, precision, scale, size)
+    else:
+        unscaled_datum = read_decimal_from_fixed2(data,
+                                                  precision, scale, size)
+
+    original_prec = getcontext().prec
```
tebeka: Use [localcontext](https://docs.python.org/2/library/decimal.html#decimal.localcontext) as context manager.

fastavro/reader.py
```diff
@@ -134,6 +142,87 @@ def read_boolean(fo, writer_schema=None, reader_schema=None):
...
+    number of bytes declared in the schema.
+    based on https://github.com/apache/avro/pull/82/
+    """
+    unscaled_datum = 0
+    msb = datum[0]
+    leftmost_bit = (msb >> 7) & 1
+    if leftmost_bit == 1:
+        modified_first_byte = (datum[0]) ^ (1 << 7)
+        datum = (modified_first_byte).to_bytes(1, 'big') + datum[1:]
+        for offset in range(size):
```
tebeka: Use xrange

fastavro/reader.py
```diff
@@ -134,6 +142,87 @@ def read_boolean(fo, writer_schema=None, reader_schema=None):
...
+            unscaled_datum <<= 8
+            unscaled_datum += datum[offset]
+        unscaled_datum += pow(-2, (size * 8) - 1)
+    else:
+        for offset in range(size):
```
tebeka: Use xrange

fastavro/reader.py
```diff
@@ -134,6 +142,87 @@ def read_boolean(fo, writer_schema=None, reader_schema=None):
...
+    Decimal is encoded as fixed. Fixed instances are encoded using the
+    number of bytes declared in the schema.
+    based on https://github.com/apache/avro/pull/82/
+    """
+    unscaled_datum = 0
+    msb = struct.unpack('!b', datum[0])[0]
+    leftmost_bit = (msb >> 7) & 1
+    if leftmost_bit == 1:
+        modified_first_byte = ord(datum[0]) ^ (1 << 7)
+        datum = chr(modified_first_byte) + datum[1:]
```
tebeka: Is the only line that differs from `read_decimal_from_fixed`? If so maybe we can find something that work on both version.

fastavro/reader.py
```diff
@@ -375,10 +471,17 @@ def read_data(fo, writer_schema, reader_schema=None):
     """Read data from file object according to schema."""
 
     record_type = extract_record_type(writer_schema)
+    logical_type = extract_logical_type(writer_schema)
```
tebeka: No need for new variable. IMO change `extract_record_type` to return the logical type.

fastavro/schema.py
```diff
@@ -42,6 +42,15 @@ def extract_record_type(schema):
     return schema
 
 
+def extract_logical_type(schema):
+    if isinstance(schema, dict):
```
tebeka: Make the code a bit more explicit.
```python
    if not isinstance(schema, dict):
        return None
   lt = schema.get('logicalType)
   if not lt:
        return None
    return rt + '-' + lt
```

fastavro/writer.py
```diff
@@ -42,6 +44,68 @@ def write_boolean(fo, datum, schema=None):
...
 
 
+def write_timestamp_millis(data, schema):
+    t = int(time.mktime(data.timetuple())) * 1000 + int(
+        data.microsecond / 1000)
+    return t
+    # return int(data.timestamp() * 1000)
+
+
+def write_timestamp_micros(data, schema):
```
tebeka: Don't call these `write_` since they don't write. Maybe `_to_logical`?

fastavro/writer.py
```diff
@@ -42,6 +44,68 @@ def write_boolean(fo, datum, schema=None):
...
+    t = int(time.mktime(data.timetuple())) * 1000000 + data.microsecond
+    return t
+    # return int(data.timestamp() * 100000)
```
tebeka: Remove commented lines

fastavro/writer.py
```diff
@@ -42,6 +44,68 @@ def write_boolean(fo, datum, schema=None):
...
+    padding_bits = ~((1 << bits_req) - 1) if sign else 0
+    packed_bits = padding_bits | unscaled_datum
+
+    bytes_req += 1 if (bytes_req << 3) < bits_req else 0
+
+    tmp = MemoryIO()
+
+    for index in range(bytes_req - 1, -1, -1):
+        bits_to_write = packed_bits >> (8 * index)
+        if PY3:
```
tebeka: Do this at definition time. Maybe add a function to `six.py`?

fastavro/reader.py
kkirsanov: no time for test just now. So i`l return the "fastavfo"

fastavro/writer.py
kkirsanov: done

fastavro/writer.py
kkirsanov: A bit complicated, but i'll try

fastavro/writer.py
kkirsanov: What about "prepare" or "serialize"?
prepare_timestamp::datetime->int
prepare_bytes_decimal::decimall->bytes

fastavro/schema.py
kkirsanov: done

fastavro/reader.py
kkirsanov: done

fastavro/reader.py
kkirsanov: done

fastavro/reader.py
kkirsanov: done

fastavro/reader.py
kkirsanov: done

fastavro/reader.py
kkirsanov: done

fastavro/reader.py
kkirsanov: In this case we have to pot two different classes of parses into one dispatcher. Keeping them separated makes things simpler.

fastavro/reader.py
kkirsanov: Agree. I'll do it today

tebeka:
Thanks! Will have a look this weekend.

kkirsanov:
About perfomance.
on this test case
https://gist.github.com/kkirsanov/8f685a4c3992bd8ff3065b092b3d7773

Without cpython  LogicalTypes version are 1.0997547 times slower

PR #80, opened 2017-06-07
leftmost bit decimal bug

kkirsanov:
During the extended testing on the production data, I found an error in python 3+ :(

Sorry for the lack of testing.


tebeka:
Thanks!

PR #88, opened 2017-08-01
Fix incorrect verification of datetime logical type

kkirsanov:
This bug appears when unions are used.

Like a {"type": ["null", {"type": "long", "logicalType": "timestamp-micros"}]}


kkirsanov:
Sory.

Werong history in git.

PR #89, opened 2017-08-01
Fix incorrect verification of datetime logical type

kkirsanov:
This bug appears when unions are used. Like a {"type": ["null", {"type": "long", "logicalType": "timestamp-micros"}]}

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+2017-08-1 version 0.14.6
+ * Fix incorrect verification of datetime logical type (@kkirsanov in PR #89)
+
 2017-07-15 version 0.14.5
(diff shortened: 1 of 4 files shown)


kkirsanov:
For the future.
Have I done the right thing before merging?

tebeka:
> Have I done the right thing before merging?
I don't understand the question? If you mean submit a PR then probably yes. Unless it's a trivial fix I prefer people submit a PR and let other opportunity to comment on it. After few days you can merge it if there are no objections.

kkirsanov:
Did I make all the necessary changes (version in __init__.py,  Changelog etc.) to merge it myself, without your attention.

tebeka:
IMO yes. Maybe we can add a `HACKING.md` file to explain.

PR #91, opened 2017-08-04
Update writer.py

luup2k:
typo on long validator function using INT constant intead long constants
**Side note**: this behaviour only happens when using write_union function 

Test Case:

```
from fastavro import writer

schema = {
    'doc': 'A weather reading.',
    'name': 'Weather',
    'namespace': 'test',
    'type': 'record',
    'fields': [
        {'name': 'time', 'type': ['null', 'long' ] },
    ],
}

records = [
    { u'time': 1433270389},
    { u'time': 809066167221092352},
]

with open('weather.avro', 'wb') as out:
    writer(out, schema, records)
```

### DIFF
--- a/tests/test_fastavro.py
+++ b/tests/test_fastavro.py
@@ -773,3 +773,27 @@ def test_is_avro_fo():
             assert fastavro.is_avro(fp)
     with open(__file__, 'rb') as fp:
         assert not fastavro.is_avro(fp)
+
+
+def test_write_long_union_type():
+    schema = {
+        'name': 'test_name',
+        'namespace': 'test',
+        'type': 'record',
+        'fields': [
+            {'name': 'time', 'type': ['null', 'long' ] },
+        ],
+    }
+
+    new_file = MemoryIO()
+    records = [ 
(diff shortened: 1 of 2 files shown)


tebeka:
Thanks! Can you please add a test that demonstrates the problem?

luup2k:
Test added on test_fastavro.py: test_write_long_union_type

tests/test_fastavro.py
```diff
@@ -773,3 +773,27 @@ def test_is_avro_fo():
...
+            {'name': 'time', 'type': ['null', 'long']},
+        ],
+    }
+
+    new_file = MemoryIO()
+    records = [
+        {u'time': 809066167221092352},
+    ]
+    try:
+        fastavro.writer(new_file, schema, records)
```
scottbelden: I would remove the try/except and just call the writer. An exception will cause a failure just like the `assert False`. Unless there is something else going on here that I'm missing...

luup2k:
try/except removed from test_fastavro.py

scottbelden:
Thanks! Looks good to me.

PR #92, opened 2017-08-06
Fix deserializing logicalType timestamp-millis millisecond gotcha

dodysw:
Fix deserializing logicalType timestamp-millis was storing millisecond  inside Python's datatime microsecond property.

### DIFF
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -141,16 +141,15 @@ def read_boolean(fo, writer_schema=None, reader_schema=None):
 
 
 def parse_timestamp(data, resolution):
-    return datetime.datetime.fromtimestamp(data / resolution).replace(
-        microsecond=data % resolution)
+    return datetime.datetime.fromtimestamp(data / resolution)
 
 
 def read_timestamp_millis(data, writer_schema=None, reader_schema=None):
-    return parse_timestamp(data, 1000)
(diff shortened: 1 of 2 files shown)


fastavro/reader.py
```diff
@@ -141,16 +141,15 @@ def read_boolean(fo, writer_schema=None, reader_schema=None):
...
 
 def parse_timestamp(data, resolution):
-    return datetime.datetime.fromtimestamp(data / resolution).replace(
-        microsecond=data % resolution)
+    return datetime.datetime.fromtimestamp(data / resolution)
 
 
 def read_timestamp_millis(data, writer_schema=None, reader_schema=None):
-    return parse_timestamp(data, 1000)
+    return parse_timestamp(data, 1000.0)
```
scottbelden: Why do these need to be floats? Otherwise this looks good to me.

dodysw: Unlike Python 3, In Python 2 division operator (at parse_timestamp) by integer will output integer and loose fractional seconds.. The float here forces division operator to do a float division in Python 2, and not impacting Python 3.

scottbelden:
Looks good to me!

tebeka:
@scottbelden Feel free to merge if you're OK with it (next time ...)

scottbelden:
Will do.

Issue #105, opened 2017-10-27
Add support for fixed type for logicalType decimals


glgnohk:
I.e.: 
```
  "type": 
    {
        "name": "MyFixedDecimal",
        "type": "fixed",
        "size": 8,
        "logicalType": "decimal",
        "precision": 18,
        "scale": 6
    },
```

I think what has to be done:

### Add "fixed-decimal" to LOGICAL_READERS and LOGICAL_WRITERS
"bytes-decimal" is in both of these, but not "fixed-decimal"

### Add functions to the above
For the reader, function ```read_bytes_decimal``` should be able to handle the fixed-decimals, as it does ```size = len(data)```.

For the writer, will need to make a new function. Can take from https://github.com/apache/avro/pull/82/, like the current code.

### Add "fixed-decimal" to validate
Add:
```
if record_type == 'fixed':
        return (
            (isinstance(datum, bytes) and len(datum) == schema['size'])
            or (isinstance(datum, decimal.Decimal))
        )
```

Union calls validate, so you'll need this if use union with "fixed-decimal".

**You could monkey patch the reader with:**
```
fastavro._reader.LOGICAL_READERS['fixed-decimal']=fastavro._reader.LOGICAL_READERS['bytes-decimal']
```

I could attempt a PR.

scottbelden:
This sounds like it shouldn't be too hard to implement. Feel free to submit a PR. Thanks!

scottbelden:
resolved via https://github.com/tebeka/fastavro/pull/106

PR #106, opened 2017-10-30
Added fixed decimal type (fix for #105)

glgnohk:
Some comments:

*  In ```tests/test_logical_types.py```, why was the schema for bytes decimal called ```schema_top```? There's also a test called ```test_top()```, which was just a copy of ```test_bytes_decimal()```. Can we rename the schema  ```schema_fixed_decimal``` and get rid of ```test_top()```?
* In ```reader.py```, the comment of function ```read_bytes_decimal()``` sounded like it was for fixed decimals, so I moved this comment to the new function ``` read_fixed_decimal()```, and reworded the comment in ```read_bytes_decimal()```.

Tested in 
* ```Python 3.5.2```
* ```Python 2.7.12```

FYI
.cache directory from pytest should probably be in .gitignore.

Happy to update based on comments.

### DIFF
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -182,14 +182,46 @@ def read_time_micros(data, writer_schema=None, reader_schema=None):
 
 def read_bytes_decimal(data, writer_schema=None, reader_schema=None):
     """
-    Decimal is encoded as fixed. Fixed instances are encoded using the
-    number of bytes declared in the schema.
+    Decimal is encoded as bytes.
     based on https://github.com/apache/avro/pull/82/
     """
     scale = writer_schema['scale']
     precision = writer_schema['precision']
 
     size = len(data)
+
+    datum_byte = str2ints(data)
+
+    unscaled_datum = 0
+    msb = fstint(data)
+    leftmost_bit = (msb >> 7) & 1
+    if leftmost_bit == 1:
+        modified_first_byte = datum_byte[0] ^ (1 << 7)
+        datum_byte = [modified_first_byte] + datum_byte[1:]
+        for offset in xrange(size):
+            unscaled_datum <<= 8
+            unscaled_datum += datum_byte[offset]
+        unscaled_datum += pow(-2, (size * 8) - 1)
+    else:
+        for offset in xrange(size):
+            unscaled_datum <<= 8
+            unscaled_datum += (datum_byte[offset])
+
+    with localcontext() as ctx:
+        ctx.prec = precision
+        scaled_datum = Decimal(unscaled_datum).scaleb(-scale)
+    return scaled_datum
+
+
+def read_fixed_decimal(data, writer_schema=None, reader_schema=None):
+    """
+    Decimal is encoded as fixed. Fixed instances are encoded using the
+    number of bytes declared in the schema.
+    based on https://github.com/apache/avro/pull/82/
+    """
+    scale = writer_schema['scale']
+    precision = writer_schema['precision']
+    size = writer_schema['size']
 
     datum_byte = str2ints(data)
 
@@ -435,6 +467,7 @@ LOGICAL_READERS = {
     'long-timestamp-micros': read_timestamp_micros,
     'int-date': read_date,
     'bytes-decimal': read_bytes_decimal,
+    'fixed-decimal': read_fixed_decimal,
     'string-uuid': read_uuid,
     'int-time-millis': read_time_millis,
     'long-time-micros': read_time_micros,
--- a/fastavro/reader.py
+++ b/fastavro/reader.py
@@ -181,47 +181,21 @@ def read_time_micros(data, writer_schema=None, reader_schema=None):
 
 
 def read_bytes_decimal(data, writer_schema=None, reader_schema=None):
-    """
-    Decimal is encoded as bytes.
-    based on https://github.com/apache/avro/pull/82/
-    """
-    scale = writer_schema['scale']
-    precision = writer_schema['precision']
-
     size = len(data)
-
-    datum_byte = str2ints(data)
-
-    unscaled_datum = 0
-    msb = fstint(data)
-    leftmost_bit = (msb >> 7) & 1
-    if leftmost_bit == 1:
-        modified_first_byte = datum_byte[0] ^ (1 << 7)
-        datum_byte = [modified_first_byte] + datum_byte[1:]
-        for offset in xrange(size):
-            unscaled_datum <<= 8
-            unscaled_datum += datum_byte[offset]
-        unscaled_datum += pow(-2, (size * 8) - 1)
-    else:
-        for offset in xrange(size):
-            unscaled_datum <<= 8
-            unscaled_datum += (datum_byte[offset])
-
-    with localcontext() as ctx:
-        ctx.prec = precision
-        scaled_datum = Decimal(unscaled_datum).scaleb(-scale)
-    return scaled_datum
+    return _read_decimal(data, size, writer_schema)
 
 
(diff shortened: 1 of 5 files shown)


fastavro/reader.py
```diff
@@ -214,6 +213,39 @@ def read_bytes_decimal(data, writer_schema=None, reader_schema=None):
     return scaled_datum
 
 
+def read_fixed_decimal(data, writer_schema=None, reader_schema=None):
```
scottbelden: The only difference I see between this and `read_bytes_decimal` is `size = writer_schema['size']` vs `size = len(data)` (unless I'm missing something). Is there a way to define `size` that works for both and have the `LOGICAL_READERS` dictionary map both `bytes-decimal` and `fixed-decimal` to the same function?

fastavro/writer.py
```diff
@@ -139,6 +139,67 @@ def prepare_bytes_decimal(data, schema):
...
+        return data
+    scale = schema['scale']
+    size = schema['size']
+
+    # based on https://github.com/apache/avro/pull/82/
+
+    sign, digits, exp = data.as_tuple()
+
+    if -exp > scale:
+        raise AssertionError(
```
scottbelden: Do `AssertionError`s get optimized away in the same way that the `assert` statement does? If so, it might be better to raise a `ValueError` or something that won't be optimized away.

fastavro/writer.py
```diff
@@ -139,6 +139,67 @@ def prepare_bytes_decimal(data, schema):
     return tmp.getvalue()
 
 
+def prepare_fixed_decimal(data, schema):
```
scottbelden: Same questions about this and `prepare_bytes_decimal`. Do we need both functions? Is this one different?

fastavro/reader.py
glgnohk: That is the only difference. I didn't like doing this, but didn't see an alternative without breaking the current coding patterns.

For example, we could do:
```
def read_fixed_decimal(data, writer_schema=None, reader_schema=None, size):
   ...

def read_bytes_decimal(data, writer_schema=None, reader_schema=None):
   size = len(data)
   read_fixed_decimal(data, writer_schema, reader_schema, size)
```

but adding the ```size``` param doesn't fit in with the other ```read_X``` functions.

Alternatively, we could use the current ``` read_bytes_decimal()``` for both bytes and fixed. It would work, but it if fixed got passed 6 bytes instead of 8 in the schema, say, it would still parse it .


I suppose we could do:
```
def read_decimal(data, writer_schema=None, reader_schema=None):
   if writer_schema:
      if writer_schema['type'] == 'fixed':
         size = writer_schema['size']
      elif  writer_schema['type'] == 'bytes
         size = len(data)
   ...
```

fastavro/writer.py
glgnohk: This one is a bit different to the reading, as there's more differences in the code.

Common code for both ```prepare_fixed_decimal``` and  ```prepare_bytes_decimal```:
```
    if not isinstance(data, decimal.Decimal):
        return data
    scale = schema['scale']

    # based on https://github.com/apache/avro/pull/82/

    sign, digits, exp = data.as_tuple()

    if -exp > scale:
        raise AssertionError(
            'Scale provided in schema does not match the decimal')
    delta = exp + scale
    if delta > 0:
        digits = digits + (0,) * delta

    unscaled_datum = 0
    for digit in digits:
        unscaled_datum = (unscaled_datum * 10) + digit

    # 2.6 support
    if not hasattr(unscaled_datum, 'bit_length'):
        bits_req = len(bin(abs(unscaled_datum))) - 2
    else:
        bits_req = unscaled_datum.bit_length() + 1
```

But the rest can be pretty different between bytes and fixed. It's not the same as reading, where only the size param is different and the rest of the calc is exactly the same. There's a much longer code path if you need to include size.  I suppose we could do a couple of 
```
      if writer_schema['type'] == 'fixed':
         ... do this whole section of code
      elif  writer_schema['type'] == 'bytes
         ... do this other whole section of code
```
But maybe different enough to have different functions.

I'm not a bit manipulation expert though, so perhaps there's better ways. Having separate functions also paves the way for a bit manipuation expert to more easily change the code.

Am happy to update on request of this repo's maintainers though.

fastavro/writer.py
glgnohk: FYI, I used the code from https://github.com/apache/avro/pull/82/ as a guide and matched the coding style of the existing ```  prepare_bytes_decimal```, which included this AssertionError. I'm a newbie to the repo and didn't want to change too much. :-)

Anyway, regardless of it "optimizing away", it's probably not the correct error.  The consensus seems to be that assertions are for debugging, and not for bad use of an API.

The avro code uses
```
raise AvroTypeException('Scale provided in schema does not match the decimal')
```

I don't think we need a new error. ```ValueError``` sounds good.

glgnohk:
@scottbelden, and thoughts on:

> In tests/test_logical_types.py, why was the schema for bytes decimal called schema_top? There's also a test called test_top(), which was just a copy of test_bytes_decimal(). Can we rename the schema schema_fixed_decimal and get rid of test_top()?

Just a quick cleanup.

fastavro/reader.py
scottbelden: What about something like this:

```python
def read_bytes_decimal(data, writer_schema=None, reader_schema=None):
    size = len(data)
    return _read_decimal(data, size, writer_schema)

def read_fixed_decimal(data, writer_schema=None, reader_schema=None):
    size = writer_schema['size']
    return _read_decimal(data, size, writer_schema)

def _read_decimal(data, size, writer_schema):
    scale = writer_schema['scale']
    precision = writer_schema['precision']

    datum_byte = str2ints(data)

    unscaled_datum = 0
    msb = fstint(data)
    leftmost_bit = (msb >> 7) & 1
    if leftmost_bit == 1:
        modified_first_byte = datum_byte[0] ^ (1 << 7)
        datum_byte = [modified_first_byte] + datum_byte[1:]
        for offset in xrange(size):
            unscaled_datum <<= 8
            unscaled_datum += datum_byte[offset]
        unscaled_datum += pow(-2, (size * 8) - 1)
    else:
        for offset in xrange(size):
            unscaled_datum <<= 8
            unscaled_datum += (datum_byte[offset])

    with localcontext() as ctx:
        ctx.prec = precision
        scaled_datum = Decimal(unscaled_datum).scaleb(-scale)
    return scaled_datum
```

It adds a function call, but reduces the duplication. Thoughts?

fastavro/writer.py
scottbelden: Okay, makes sense. I suppose we could factor the common piece out, but it's probably fine as is for the moment.

scottbelden:
@kkirsanov Would might be able to give more insight to the naming. I'm guessing it was probably just a copy/paste or an initial name and then never renamed. I think renaming the schema/test to something more descriptive sounds like a great idea 👍

fastavro/reader.py
glgnohk: Looks good to me.

fastavro/writer.py
glgnohk: Updated, including unit tests.

glgnohk:
Added all the changes suggested, even though there could be more discussion. Can change later if needed. Looks like #110 waiting on this.

I see conflicts, but #110 says we'll merge this PR first. Conflict resolution looks simple.

scottbelden:
I went ahead and resolved the merge conflict and merged it in. Thanks! @tebeka Do you mind releasing `0.14.11` whenever you have time?

Issue #87, opened 2017-08-01
Add support for aliases


regisb:
Currently, fastavro does not support Avro aliases: http://avro.apache.org/docs/current/spec.html#Aliases

Aliases are required for renaming fields, which is a frequent type of schema evolution. At read time, when a field is absent from the writer schema, the field name is looked up among the field aliases.

If you agree this is a feature worth having, I can start working on a PR.

EDIT: simplified the description of this issue to reduce the scope of aliases.

yiga2:
I second this request - useful if one maintains technical & "business" field names. It is not clear to me if **doc** attribute is supported at field-level either. 
If not, @regisb I like to piggyback on your PR as adding support for both attributes would be in the same place.

regisb:
@yiga2 I'm happy to see some interest in this feature! I'm going to be too busy to work on it in the next couple weeks, but I should be able to open a PR before the end of the year. I'm not sure if `doc` is already supported; if not, I will add it to the PR.

scottbelden:
Feel free to submit a PR. Thanks!

regisb:
@yiga2 I just opened a PR to implement aliases, but I did not work on `doc` attributes. It seems to me that they are supported. Which use case did you have in mind?

yiga2:
To store more descriptive information about the record (table) and
column/field. Aliases are useful as well to have a technical and business
name for a field but have a long text to explain further (=doc attribute)
is even better as we are dealing with many flavors of the same data,
depending on the customer. Maintaining a separate data dictionary would be
unnecessary.

I appreciate you reaching out but I will give spavro a try or even avsc.
The latter is for node.js but is superfast, which is critical in my use
case: encoding of 2GB csv files in less than 5min as I want to use
serverless functions (AWS Lambda or G/Az Functions).

Thanks anyways.

scottbelden:
Resolved via https://github.com/tebeka/fastavro/pull/113

PR #113, opened 2017-11-16
Add support for Avro aliases

regisb:
Aliases allow a writer field to have a different name than the reader
field. Potentially, there may be multiple reader fields pointing to the
same writer field, and this is the biggest difficulty to implement
aliases. Fields are read only once from file, but they may need to be
validated multiple times. Files may be read only once because we do not
support "seek()" statements.

The approach we chose was to create a file reader with a cache. A
potential issue with this is for aliased fields that take a lot of
space. We try to mitigate this issue by resorting to in-memory storage
as infrequently as possible.

This should close issue #87.

EDIT: we no longer try to handle the case when multiple reader fields point to the same writer field.

### DIFF
--- a/tests/test_fastavro.py
+++ b/tests/test_fastavro.py
@@ -601,6 +601,56 @@ def test_schema_migration_reader_union():
     assert new_records == records
 
 
+def test_aliases_not_present():
+    schema = {
+        "type": "record",
+        "fields": [{
+            "name": "test",
+            "type": "double"
+        }]
+    }
+
+    new_schema = {
+        "type": "record",
+        "fields": [
+            {"name": "newtest", "type": "double", "aliases": ["testX"]},
+        ]
+    }
+
+    new_file = MemoryIO()
+    records = [{"test": 1.2}]
+    fastavro.writer(new_file, schema, records)
+    new_file.seek(0)
+    reader = fastavro.reader(new_file, new_schema)
+    with pytest.raises(fastavro.read.SchemaResolutionError):
+        list(reader)
+
+
+def test_incompatible_aliases():
+    schema = {
+        "type": "record",
+        "fields": [{
+            "name": "test",
+            "type": "double"
+        }]
+    }
+
+    new_schema = {
+        "type": "record",
+        "fields": [
+            {"name": "newtest", "type": "int", "aliases": ["test"]},
+        ]
+    }
+
+    new_file = MemoryIO()
+    records = [{"test": 1.2}]
+    fastavro.writer(new_file, schema, records)
+    new_file.seek(0)
+    reader = fastavro.reader(new_file, new_schema)
+    with pytest.raises(fastavro.read.SchemaResolutionError):
+        list(reader)
+
+
 def test_schema_migration_union_failure():
     schema = {
         "type": "record",
@@ -737,6 +787,33 @@ def test_schema_migration_schema_mismatch():
     new_reader = fastavro.reader(new_file, new_schema)
     with pytest.raises(fastavro.read.SchemaResolutionError):
         list(new_reader)
+
+
+def test_aliases_in_reader_schema():
+    schema = {
+        "type": "record",
+        "fields": [{
+            "name": "test",
+            "type": "int"
+        }]
+    }
+
+    new_schema = {
+        "type": "record",
+        "fields": [{
+            "name": "newtest",
+            "type": "int",
+            "aliases": ["test"]
+        }]
+    }
+
+    new_file = MemoryIO()
+    records = [{"test": 1}]
+    fastavro.writer(new_file, schema, records)
+    new_file.seek(0)
+    new_reader = fastavro.reader(new_file, new_schema)
+    new_records = list(new_reader)
+    assert new_records[0]["newtest"] == records[0]["test"]
 
 
 def test_empty():
--- a/fastavro/_read.pyx
+++ b/fastavro/_read.pyx
@@ -454,13 +454,21 @@ cpdef read_record(ReaderBase fo, writer_schema, reader_schema=None):
         for field in writer_schema['fields']:
             record[field['name']] = _read_data(fo, field['type'])
     else:
-        readers_field_dict = {f['name']: f for f in reader_schema['fields']}
+        readers_field_dict = {}
+        aliases_field_dict = {}
+        for f in reader_schema['fields']:
+            readers_field_dict[f['name']] = f
+            for alias in f.get('aliases', []):
+                aliases_field_dict[alias] = f
+
         for field in writer_schema['fields']:
             readers_field = readers_field_dict.get(field['name'])
+            if not readers_field:
+                readers_field = aliases_field_dict.get(field['name'])
             if readers_field:
-                record[field['name']] = _read_data(fo,
-                                                   field['type'],
-                                                   readers_field['type'])
+                record[readers_field['name']] = _read_data(fo,
+                                                           field['type'],
+                                                           readers_field['type'])
             else:
                 # should implement skip
                 _read_data(fo, field['type'], field['type'])


scottbelden:
Thanks for the PR! I'll try to take a look tomorrow or over the weekend.

tests/test_fastavro.py
```diff
@@ -595,6 +596,119 @@ def test_schema_migration_reader_union():
...
+        }]
+    }
+
+    new_file = MemoryIO()
+    records = [{"test1": 1, "test2": 2}]
+    fastavro.writer(new_file, schema, records)
+    new_file.seek(0)
+    new_reader = fastavro.reader(new_file, new_schema)
+    new_records = list(new_reader)
+    assert new_records[0]["test2"] == records[0]["test2"]
```
scottbelden: I'm not sure about this test. The specification for aliases states `Aliases function by re-writing the writer's schema using aliases from the reader's schema. For example, if the writer's schema was named "Foo" and the reader's schema is named "Bar" and has an alias of "Foo", then the implementation would act as though "Foo" were named "Bar" when reading.`

That seems to imply to me that we should treat the original schema's `test1` field as being named `test2` (and we would ignore the original schema's `test2` field) so the assertion should be `assert new_records[0]["test2"] == records[0]["test1"]`

What are your thoughts?

tests/test_fastavro.py
```diff
@@ -595,6 +596,119 @@ def test_schema_migration_reader_union():
...
+
+
+def test_same_alias_for_multiple_fields():
```
scottbelden: I think this test case and feature is the reason for the `rewindable_reader`. I'm curious to know if this is a real scenario you experience right now. It seems strange to me that two fields on a new schema would alias the same old field...

tests/test_fastavro.py
regisb: I must admit that I slightly interpreted the documentation: the alias is used only if a field with the original name is not present. This is confirmed by the implementation of the `avro` python package:

 ```
import avro.schema
from avro.datafile import DataFileReader, DataFileWriter
from avro.io import DatumReader, DatumWriter

# Write data
writers_schema = avro.schema.make_avsc_object({
    "type": "record",
    "name": "pouac",
    "fields": [
        {
            "name": "test1",
            "type": "int"
        },
        {
            "name": "test2",
            "type": "int"
        }
    ]
})
writer = DataFileWriter(open("records.avro", "wb"), DatumWriter(), writers_schema)
writer.append({"test1": 1, "test2": 2})
writer.close()

# Read data
readers_schema = avro.schema.make_avsc_object({
    "type": "record",
    "name": "pouac",
    "fields": [{
        "name": "test2",
         "type": "int",
         "aliases": ["test1"]
    }]
})
reader = DataFileReader(open("records.avro", "rb"), DatumReader(readers_schema=readers_schema))
for user in reader:
    print user
reader.close()
```

This script prints `{u'test2': 2}`, so the `avro` package has the same behaviour.

File: tests/test_fastavro.py (hunk 2 above)
Comment (regisb) [reply tests/test_fastavro.py]: Yes, exactly. We would not need the (admittedly hackish) `rewindable_reader` if we did not have to validate data against multiple types.
You could decide to support aliases, but not multiple aliases to the same name. The current implementation would be very much simplified. Having multiple fields point to the same aliased field has never occurred to me, but I can imagine how it could happen: when a field becomes superseded by another, it is still used downstream for some time. It might be important to provide correct values for this legacy field, at least for some time.

**Comment by regisb** [comment]
I'm confused by the new  `_reader.pyx` and `_reader_py.py` files? Do I need to add my changes to both files?

File: tests/test_fastavro.py (hunk 1 above)
Comment (scottbelden) [reply tests/test_fastavro.py]: The python version of the standard `avro` package doesn't implement aliasing (https://issues.apache.org/jira/browse/AVRO-1303). I think I read that the java code does implement it, so it would be interesting to see what happens on the java side.

File: tests/test_fastavro.py (hunk 2 above)
Comment (scottbelden) [reply tests/test_fastavro.py]: If you don't have a need for it then I think it's best to wait until there is a need. It can be implemented at that time and until then it will make things simpler.

**Comment by scottbelden** [comment]
Yes, hopefully that should be too hard, but I think we should resolve the open questions first before committing time to updating both files

File: tests/test_fastavro.py (hunk 2 above)
Comment (regisb) [reply tests/test_fastavro.py]: I'm fine with that.

File: tests/test_fastavro.py (hunk 1 above)
Comment (regisb) [reply tests/test_fastavro.py]: HA! I didn't know that. I'll summon old-forgotten Java skills and check the Java implementation.

**Comment by barrywhart** [comment]
@regisb: I would focus on updating the `py` files first. The `.pyx` files are pretty similar in structure. There are two ways to use the `.py` files:
* With Pypy (as opposed to CPython), the `.py`  files are used automatically.
* If you prefer to use CPython for development, temporarily replace the code that imports the `.pyx` files so it uses the `.py` files instead. (Don't commit these changes, though.)

Once you have working `.py` files, we can worry about updating the `.pyx` files.

**Comment by nfisher** [comment]
Might it be worth writing a schema validator class similar to Java? It would allow for simple CLI validation of two distinct schemas as well as being used as a "fail-fast" validation prior to processing any records in an Avro file.

**Comment by scottbelden** [comment]
@nfisher That sounds like it would be useful if it were a standalone utility/function and would also resolve https://github.com/tebeka/fastavro/issues/136.

**Comment by SimonShapiro** [comment]
8Hi,

I am tracking this PR because I need the functionality described.  At the moment I have a work around that re-writes the returned dictionary for each record according to the alias functionality described in the  avro spec.  It inspects the writers schema and applies aliases where they exist or defaults everywhere that is not matched between the readers and writers schemas.  

Would you be interested in seeing that function?

**Comment by scottbelden** [comment]
@SimonShapiro If you'd like to submit a PR feel free to. I personally haven't had a need for this and so I haven't pressed to get this in, but if you have something that is most of the way there then we can work to get that merged in.

**Comment by regisb** [comment]
I checked the reference Java implementation of Avro to see what is the correct behavior to adopt in corner cases. Here's the Java code: 

```
package testavro;

import java.io.File;
import java.io.IOException;

import org.apache.avro.Schema;
import org.apache.avro.file.DataFileReader;
import org.apache.avro.file.DataFileWriter;
import org.apache.avro.generic.GenericData;
import org.apache.avro.generic.GenericDatumReader;
import org.apache.avro.generic.GenericDatumWriter;
import org.apache.avro.generic.GenericRecord;
import org.apache.avro.io.DatumReader;
import org.apache.avro.io.DatumWriter;

public class TestAvro {

	public static void main(String[] args) throws IOException {
		// Load schemas
		Schema writerSchema = new Schema.Parser().parse(new File("/home/regis/avro/writer.avsc"));
		Schema readerSchema = new Schema.Parser().parse(new File("/home/regis/avro/reader.avsc"));

		// Create record
		GenericRecord pouac1 = new GenericData.Record(writerSchema);
		pouac1.put("test1", 1);
		pouac1.put("test2", 2);
		
		// Save record
		File fileWrite = new File("/home/regis/avro/records.avro");
		DatumWriter<GenericRecord> datumWriter = new GenericDatumWriter<GenericRecord>(writerSchema);
		DataFileWriter<GenericRecord> dataFileWriter = new DataFileWriter<GenericRecord>(datumWriter);
		dataFileWriter.create(writerSchema, fileWrite);
		dataFileWriter.append(pouac1);
		dataFileWriter.close();
		
		// Read record
		File fileRead = new File("/home/regis/avro/records.avro");
		DatumReader<GenericRecord> datumReader = new GenericDatumReader<GenericRecord>(readerSchema);
		DataFileReader<GenericRecord> dataFileReader = new DataFileReader<GenericRecord>(fileRead, datumReader);
		GenericRecord pouac = null;
		System.out.println(dataFileReader.next(pouac));
		dataFileReader.close();
	}
}
```

`writer.avsc`:

```
{
    "type": "record",
    "name": "pouac",
    "fields": [
        {
            "name": "test1",
            "type": "int"
        },
        {
            "name": "test2",
            "type": "int"
        }
    ]
}
```

`reader.avsc`:

```
{
    "type": "record",
    "name": "pouac",
    "fields": [{
        "name": "test2",
         "type": "int",
         "aliases": ["test1"]
    }]
}
```

The code above crashed with an AvroRuntimeException: 

```
Exception in thread "main" org.apache.avro.AvroRuntimeException: Duplicate field test2 in record pouac: test2 type:INT pos:1 and test2 type:INT pos:0.
	at org.apache.avro.Schema$RecordSchema.setFields(Schema.java:651)
	at org.apache.avro.Schema.applyAliases(Schema.java:1442)
	at org.apache.avro.Schema.applyAliases(Schema.java:1418)
	at org.apache.avro.generic.GenericDatumReader.getResolver(GenericDatumReader.java:128)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:143)
	at org.apache.avro.file.DataFileStream.next(DataFileStream.java:233)
	at testavro.TestAvro.main(TestAvro.java:41)
```

In other words, the corner cases that I try to cover in this PR are *not* handled by the reference Java implementation.

So I'm back to the drawing board to produce a simpler implementation. Yay!

tests/test_fastavro.py
regisb: I can confirm that the reference Java implementation does *not* support multiple fields with the same alias. I'm getting a `org.apache.avro.AvroTypeException` whenever I try to read data with multiple fields that share a common alias. This is good news (sorta) because it should allow me to really simplify the code.

regisb:
Please check the new implementation -- drastically simplified.

scottbelden:
I moved the tests to their own `test_aliases.py` file and updated the `_read_py.py` so that the pypy tests should work and merged.

Thanks!

regisb:
This is great, thanks!

PR #104, opened 2017-10-23
ujson usage in case if installed

lemurchik:
There is simplejson usage in fastavro/writer.py file i think it's better to use ujson and use it not only in writer.

I'm not sure how to write about it in README.md because information about simplejson in writer wasn't reflected there.

https://github.com/esnme/ultrajson#ultrajson

Issue #36, opened 2015-10-04
Rename fastavro.writer


tebeka:
We currently have `fastavro.writer` module but also a `writer` object in `__init__.py`. This makes it hard while debugging to import the `writer` module.

ze-phyr-us:
This is a bigger problem than just debugging. When you import the reader/writer submodules, the import overwrites the `fastavro.reader` or `fastavro.writer` attributes with the imported modules:

```python
import fastavro
from fastavro.reader import SchemaResolutionError # or even just 'import fastavro.reader'

print(fastavro.reader)
```

```
$ python3 importtest.py
<module 'fastavro.reader' from 'XXX/env/lib/python3.5/site-packages/fastavro/reader.py'>
```

tebeka:
Thanks! I'll try to work on it soon-ish. Patches are always welcomed ;)

ze-phyr-us:
@tebeka I'm happy to submit a PR. What would be good names for the modules? `reading.py` and `writing.py`? `read.py` and `write.py`?

tebeka:
@ze-phyr-us Not sure. Maybe `read.py` and `write.py`
Thanks!

barrywhart:
This [PR](https://github.com/tebeka/fastavro/pull/114) addresses the issue.

scottbelden:
Resolved via https://github.com/tebeka/fastavro/pull/137

PR #114, opened 2017-11-19
Cython-based implementation of writer module (~2x faster)

barrywhart:
Inspired by the discussion on #58, I created a version of `fastavro` that takes significant advantage of Cython's capabilities:
* Replace use of `MemoryIO` with `bytearray`. Byte arrays have native support in Cython, so operations on them should be much faster.
* Add type annotations to the parameter lists of the `write_*()` functions. Also add `cdef` type declarations to some of the local variables, especially for loop variables.
* Use Cython-specific code to prepare data for serialization. In particular, eliminate use of `struct.pack()`.
* `write_data()`: For the most common data types (`boolean`, `string`, `int`, `long`, `float`, `double`), avoid using the `WRITERS` dictionary lookup, and mark the  `write_*()` functions as `inline` to avoid C function call overhead for those types.

I have not benchmarked it yet, but I believe it will perform quite a bit better than the current Cython implementation.

I did a similar Cython project a couple of years ago for decoding network packets. It used some of the techniques described above, and the final Cython version was 5-7x faster than the original, pure Python implementation. I don't expect anything quite that dramatic here, but still a meaningful improvement.

My focus in this PR was on speeding up the writer module, since the main use case on my current project involves exporting data from Python to Avro.

This PR eliminates the Python-only modules in favor of Cython (.pyx) modules, but if we don't mind code duplication, both implementations can exist side by side like in the current version.

@tebeka


barrywhart:
Tests are passing for Python 2.7, failing on other versions. I'd like to get some initial thoughts on this work before addressing the other versions.

tebeka:
We don't have a benchmark. IIRC @scottbelden is looking for something for #104 

The main reason we have the pure Python code is to support pypy and I know users who use fastavro for that reason. OTOH new version of pypy [should support Cython](http://doc.pypy.org/en/latest/release-v5.9.0.html) so if you can keep pypy support I'd love to have this.

barrywhart:
I'll try the benchmark [gist](https://gist.github.com/scottbelden/23de56f57d1b6956e3620f0bf1c3fac7) mentioned on #104.

That's interesting news about pypy. I'll see if I can try that out locally. If for some reason it's not fully compatible yet, I suppose we could keep both implementations, at least until the compatibility is better.

One thought about performance (unrelated to this PR): you could use `eval` to create Python code on the fly which is specific to each schema. This could offer a nice speed boost for pure Python, possibly the Cython version as well. (Obviously you can't generate Cython on the fly, but you could generate code on the fly which *uses* the Cython code.)

barrywhart:
I ran the benchmark script. It looks like writes are about 4x faster and reads are about 2x faster. It's interesting the writes improved so much, because I didn't make any real efforts there. Maybe they benefitted from some changes to the schema module.

Anyway, I'm pretty happy with the improvement. I'll test out the new version of pypy, then we can figure out whether to keep the existing Python modules alongside the Cython implementation.

Benchmark results:
```
This PR

Writing 1 records to one file...
... 100000 runs averaged 3.10590028763e-05 seconds
Reading 1 records from one file...
... 100000 runs averaged 2.56437039375e-05 seconds

Writing 100 records to one file...
... 1000 runs averaged 0.000138868570328 seconds
Reading 100 records from one file...
... 1000 runs averaged 0.000240780830383 seconds

... 10 runs averaged 0.0107235431671 seconds
... 10 runs averaged 0.0209925413132 seconds

Writing 1 records to one file...
... 100000 runs averaged 5.44090199471e-05 seconds
Reading 1 records from one file...
... 100000 runs averaged 6.03425192833e-05 seconds

Writing 100 records to one file...
... 1000 runs averaged 0.000689450263977 seconds
Reading 100 records from one file...
... 1000 runs averaged 0.00127932524681 seconds

... 10 runs averaged 0.0636064529419 seconds
... 10 runs averaged 0.128159594536 seconds

==

Current implementation

Writing 1 records to one file...
... 100000 runs averaged 4.93640208244e-05 seconds
Reading 1 records from one file...
... 100000 runs averaged 4.31764364243e-05 seconds

Writing 100 records to one file...
... 1000 runs averaged 0.000504611968994 seconds
Reading 100 records from one file...
... 1000 runs averaged 0.000503418684006 seconds

... 10 runs averaged 0.0454469203949 seconds
... 10 runs averaged 0.0484051942825 seconds

Writing 1 records to one file...
... 100000 runs averaged 0.000102311601639 seconds
Reading 1 records from one file...
... 100000 runs averaged 9.2787065506e-05 seconds

Writing 100 records to one file...
... 1000 runs averaged 0.00278361058235 seconds
Reading 100 records from one file...
... 1000 runs averaged 0.00273501205444 seconds

... 10 runs averaged 0.281569910049 seconds
... 10 runs averaged 0.289435243607 seconds
```

tebeka:
Thanks @barrywhart !

barrywhart:
Hmm, I'm afraid the previous benchmark numbers were incorrect, because I mistakenly ran the "current implementation" benchmark without the Cython accelerations in place. When I rerun it, the numbers are pretty similar between the two runs. I'll do some experiments to try and confirm this. I thought there'd be some speedup at least.

barrywhart:
I'm having trouble correctly running the benchmark against the master branch because I'm not sure how to build the Cython extensions in that case. It seems like `python setup.py bdist_wheel` does not build them, and if I run `python setup.py build_ext --inplace`, it fails because it tries to compile C files that do not exist. Do you know how to do this?

BTW, I think the identical benchmark numbers occurred because I ran the old (`master`) Python code using the PR's .c files, so naturally the performance was the same.

scottbelden:
Typically I build/install locally by doing `make fresh` and then `pip install .`

Also, I don't know how great the `benchmark.py` is. It was something I tried to throw together real quick as a smoke test, but it might not be an accurate representation of what should be benchmarked.

I haven't had a chance to look closely at the PR yet. I'll try to do that sometime soon.

barrywhart:
Ok, I fixed some issues with the previous benchmarks. With both versions of the code present on a machine, it's easy to accidentally run the wrong version by accident, even when using separate virtualenvs (because Python looks in the current directory for modules before it checks the virtualenv.)

So here are the correct numbers.

I ran the benchmark script with just the 10,000 record tests, comparing performance of the current release (0.14.11) against this PR. With this PR, writes are about 2x faster and reads are the same as before.

I'll test out the new version of pypy, then we can figure out whether to keep the existing Python modules alongside the Cython implementation.

Benchmark results:
```
This PR

==

0.14.11

... 10 runs averaged 0.0214914083481 seconds
... 10 runs averaged 0.0222071886063 seconds

... 10 runs averaged 0.133398509026 seconds
... 10 runs averaged 0.122091889381 seconds
```

brianbruggeman:
Problem
======

This won't install as-is.

Solution
======
Please add something like the following to setup.py:

```
    setup_requires=[
        # Setuptools 18.0 properly handles Cython extensions.
        'setuptools>=18.0',
        'cython',
    ],
```

Details
=====

This should work:

```
pip install -e git+https://github.com/barrywhart/fastavro.git@more_cython#egg=fastavro
```

But currently, if Cython is not installed, you'll see something like:
```
Obtaining fastavro from git+https://github.com/barrywhart/fastavro.git@more_cython#egg=fastavro
  Cloning https://github.com/barrywhart/fastavro.git (to more_cython) to <redacted>/.virtualenvs/avro/src/fastavro
    Complete output from command python setup.py egg_info:
      File "<redacted>/.virtualenvs/avro/src/fastavro/setup.py", line 6, in <module>
        from Cython.Build import cythonize
    ModuleNotFoundError: No module named 'Cython'
```

Reproduce steps:
==============

1.  Create and start-up a new virtual environment
2.  Run the pip install command above.

brianbruggeman:
I also did a bit of profiling, and it looks like the benchmark spends a significant (maybe 50%?) of the time running json.  We're using the standard python json library in the reader, and simplejson if and only if its installed in the writer.  I don't know why it's not consistent across both the writer and reader.  But also, there appears to be a very fast json library called [python-rapidjson](https://github.com/python-rapidjson/python-rapidjson).  My recommendation is to move to something other than the default library and potentially rapidjson (which has a number of benchmarking notes on the github page).

As a general rule, I think it would be prudent to run profiling with the benchmark to see where the hotspots are.  Otherwise you're adding cython code blindly to solve a problem you might not have.

barrywhart:
>the benchmark spends a significant (maybe 50%?) of the time running json. If that's accurate, the benchmark may be invalid.

`fastavro` only uses JSON once for each call to `reader()` or `writer()`, when processing the schema. Thus for large files, the time spent reading or writing records will dominate the run time, and the performance of the JSON library will have negligible impact.

For this reason, when I ran the benchmark, I focused on the large test cases (the ones with 10,000 records).

>adding cython code blindly to solve a problem you might not have

For this PR, I focused on write performance and an educated guess that the most commonly used types for most applications (not just the benchmark script) will be `boolean`, `string`, `int`, `long`, `float`, and `double`. I generally applied the same tuning techniques everywhere in `_writer.pyx`, because I think that solving the same problem with Python in one place and Cython in another place would have made the code harder to read.

I expect that different applications with different Avro schemas will encounter bottlenecks in different parts of the code. I spent little or no effort on certain areas, such as logical types. Those may be an area for future improvement.

barrywhart:
Pypy tests are now passing, too. I followed [these instructions](http://setuptools.readthedocs.io/en/latest/setuptools.html#distributing-extensions-compiled-with-pyrex) to update the packaging so I think it will correctly include the .c files in a source distribution and use those if Cython is not installed, but I recommend verifying this before releasing (I don't have much experience creating Python packages that use Cython).

tebeka:
Thanks @barrywhart . I'll try to have a look this week.

barrywhart:
Great, thanks! I have a few more ideas to try out for improving performance. I'll try and look at those in the next few days.

tebeka:
@barrywhart Did you benchmark pypy on pure python vs the Cython version? Last time I did that there was as slowdown.

tebeka:
@scottbelden @kkirsanov Any comments?

barrywhart:
I haven't benchmarked pypy, no. I'll try and look at that this week.

kkirsanov:
@tebeka I can look it tonight.

barrywhart:
Cython is considerably slower. I should add back the Python source files and use those for Cython, I guess?

barrywhart:
I restored the Python code. This re-surfaced some of the problems discussed in #36, so I renamed the `reader` and `writer` modules as `read` and `write`. A few tests are failing, likely due to some of the code duplicated between the `.py` and `.pyx` modules. Still looking into that...

barrywhart:
I cleaned up the module names. Now there are 3 modules for each `.py`/`.pyx` module, e.g.:
* `read.py`: Tests and other external code should use this module. It imports all the symbols from one of the two "real" modules below.
* `_read_c.pyx`: This is the Cython implementation.
* `_read_py.py`: This is the Python implementation.

Having a distinct name for each of these avoids a lot of potential confusion. It also fixes the handful of tests that were previously failing in CPython.

barrywhart:
Somehow renaming the modules affected CPython performance. Investigating...

barrywhart:
Reverted the renaming since it somehow affected performance. Made a more tactical fix.

barrywhart:
Travis won't build. Will close and re-open the PR to see if that helps.

scottbelden:
Did you mean to check in the cython generated .c files? It doesn't seem like those should be in version control if they are auto-generated.

barrywhart:
@scottbelden: Surprisingly, checking in the .c files is a [recommended practice for distributing packages with Cython extensions](http://cython.readthedocs.io/en/latest/src/reference/compilation.html#distributing-cython-modules). The rationale is that even if the destination system does not have Cython installed, it can still compile the .c files in the package. 

Of course, checking in the `.c` files is not required in order to distribute them. @tebeka: I'm not familiar with how you package and distribute the `fastavro` package. If I remove the `.c` files from git, will they still be included in the package?

barrywhart:
Actually, it looks like the `.c` files are not present in the source build, even when checked into git. I'll look into this.

barrywhart:
Ok, removed the `.c` files from git, and fixed a packaging issue that was omitting them from the source package.

scottbelden:
Why was `reader.py` changed to `_reader_py.py` but `writer.py`, `schema.py`, and `six.py` were left as originally named?

I see the 3x gain in the writer so this looks pretty promising.

.gitignore
```diff
@@ -6,7 +6,6 @@ dist
 docs/_build/
 fastavro.egg-info
 fastavro/*.so
-fastavro/_*.c
```
scottbelden: When I do a `make` or `make fresh` I still get `_*.c` files. We should keep this gitignore in or remove those files from being generated if they aren't supposed to be.

fastavro/_six.pyx
```diff
@@ -0,0 +1,99 @@
...
+    def py3_btou(n, encoding=_encoding):
+        return n.decode(encoding)
+
+    def py3_utob(n, encoding=_encoding):
+        return bytes(n, encoding)
+
+    def py3_json_dump(obj, indent):
+        json.dump(obj, stdout, indent=indent)
+
+    def py3_iterkeys(obj):
```
scottbelden: We should probably update `six.py` to have these `iterkeys` functions and update `writer.py` to use them as well (as they are used in the `pyx` files)

fastavro/_schema.pyx
```diff
@@ -0,0 +1,190 @@
...
+
+    return schema
+
+
+cpdef inline str extract_logical_type(schema):
+    cdef dict d_schema
+    if not isinstance(schema, dict):
+        return None
+    d_schema = schema
+    rt = d_schema['type']
```
scottbelden: Can you update the `schema.py` file to also do `rt = schema['type']` instead of `rt = extract_record_type(schema)` so that they are consistent?

tests/test_fastavro.py
```diff
@@ -84,21 +84,21 @@ def test_not_avro():
 
 def test_acquaint_schema_rejects_undleclared_name():
     try:
-        fastavro.schema.acquaint_schema({
+        fastavro._schema.acquaint_schema({
```
scottbelden: Did something change that makes the old imports from `fastavro.schema` not work anymore?

tests/test_schema_evolution.py
```diff
@@ -1,5 +1,8 @@
 from fastavro import writer as fastavro_writer
-from fastavro.reader import SchemaResolutionError
```
scottbelden: I feel like this could be a problem... There are probably projects that import this error: https://github.com/search?p=1&q=SchemaResolutionError&type=Code&utf8=%E2%9C%93.

We might need to add a dummy file or add another way to import this exception and mention it as a breaking change. However, if `reader.py` doesn't need to be named `_reader_py.py` then changing the name back might solve this.

.gitignore
barrywhart: I'll undo this change. It's left over from when I thought I should check in .c files.

fastavro/_schema.pyx
barrywhart: Will do.

fastavro/_six.pyx
barrywhart: Good catch! I'll update this. I made these changes in the `.pyx` files before I knew we needed to retain the `.py` files, so some things got out of sync.

barrywhart: Done.

tests/test_fastavro.py
barrywhart: I'm looking into this. A few tests fail without these changes. My hypothesis is that `schema.py` and `_schema.pyx` have mutable global state, and the breaking tests may be accidentally writing to one and reading from the other. Someone who knows this area of the code may have ideas here.

tests/test_schema_evolution.py
barrywhart: I'll see if I can avoid the need to rename. This is similar to the `schema.py` / `_schema.pyx` thing above: mistakes in the library code and/or tests can cause subtle failures and or performance problems.

barrywhart: I think I need to keep the renaming, but I created a `reader.py` with a single definition of `SchemaResolutionError` used by both `_reader_py.py` and `_reader.pyx`. Think that will work?

fastavro/_schema.pyx
barrywhart: Done

barrywhart: Done

barrywhart:
@scottbelden: I believe I've addressed all the issues you've raised. Assuming tests pass with the latest commit, this PR may be ready.

barrywhart:
>Why was reader.py changed to _reader_py.py but writer.py, schema.py, and six.py were left as originally named?

I renamed modules that had issues with the duplication of modules. Generally, modules with public classes or mutable shared state tend to have issues:
* Public classes because the two classes are not equivalent, e.g. `isinstance()` or `except SomeExceptionType` may not behave as expected, depending on which version of the symbol is imported.
* Mutable shared state: It's possible to change one copy of the state and read from the other.

I ended up having to do something similar with `schema.py`. The pattern I followed is to create a single public module, e.g. `schema.py`, to be used by clients. Related modules (e.g. `_schema_public.py`, `_schema_py.py` and `_schema.pyx`) work together to provide the public functionality.

scottbelden:
I'll try to leave some more comments when I have some time, but it seems like maybe the problem with the public/private files is that the classes should be accessed via the `fastavro/__init__.py`. Otherwise, I think what is being seen is that if you do `import fastavro.schema` you get the regular python version (need to double check on that) instead of the appropriate one.

I was playing around with some changes that would change the `_schema_py` back to `schema` and add some imports. The patch is here: https://gist.github.com/scottbelden/297fb8df1319d38b71816b8724b3fdcb. It was done quickly so I might play with it more when I have more time, but the tests pass and I think it keeps things a little cleaner.

Thoughts?

barrywhart:
Some libraries put all their public symbols in a single top-level module (e.g. `fastavro/__init__.py`) while others distribute the public interface across several modules. I usually prefer the latter, as it keeps things better organized.

It won't be possible to physically put all the symbols in `__init__.py`, because this will cause circular dependencies (e.g. modules inside fastavro need to import from `__init__.py`, but `__init__.py` in turn needs to import from them. I addressed a similar issue with `schema.py`, as described above.

`fastavro` has a pre-existing problem where `fastavro.reader` and `fastavro.writer` can be either a module or a function, depending on how you import it. This is messy and contributed to some of the problems with the tests.

barrywhart:
The patch does seem a little cleaner. The tests pass for me too, and the Cython performance is not affected. I'll commit those changes to confirm the tests pass in Travis as well.

barrywhart:
@scottbelden: I changed _reader_py.py back to reader.py, similar to your schema.py patch. I guess the concept we're evolving towards is that the Cython modules build on/override parts of the corresponding .py modules.

This will work in many cases (and seems to be fine for the current fastavro library). In the future, there could be issues with this approach, e.g. if the shared symbols from .py are not appropriate in a Cython context (e.g. if the Cython code needs a Cython version of something in order to be efficient.) We're not facing that issue yet, so let's go with the concept above for now.

scottbelden:
I agree with not using the `__init__.py` as the place for all the public symbols, but in this case it's a little different. For example, if in someone's code they do `from fastavro.reader import schemaless_reader` that would always give them the pure python version, right? However, if they do `from fastavro import schemaless_reader` they will get the cython version if available.

Of course, as you mentioned in the reader/writer/etc files we would still continue to do `from .foo import bar` instead of `from fastavro import bar`

I'll try to find some more time tomorrow to dig in a little more, but I feel like it's close.

barrywhart:
>if in someone's code they do from fastavro.reader import schemaless_reader that would always give them the pure python version, right? However, if they do from fastavro import schemaless_reader they will get the cython version if available.

Right. It's potentially confusing that these other modules do not begin with underscores, even though they are conceptually private. However, I don't want to change more than necessary, so I'm not proposing we address that now (or maybe ever).

fastavro/_reader.pyx
```diff
@@ -0,0 +1,628 @@
...
+from uuid import UUID
+
+from ._six import (
+    MemoryIO, xrange, btou, utob, iteritems, is_str, str2ints, fstint
+)
+from ._schema import (
+    extract_record_type, acquaint_schema, populate_schema_defs,
+    extract_logical_type
+)
+from .reader import SchemaResolutionError
```
scottbelden: Can we change this so that it defines the its own `SchemaResolutionError`? I feel like if we are in `_foo.pyx` files then we should only import from other `.py` files if no `.pyx` exists. I'm not sure if there are any side effects from importing the regular python file, but it just feels a little worrisome.

Alternatively we could make a `_reader_common.py` that defines the `SchemaResolutionError` and both the py and pyx reader files import from there.

fastavro/_schema_public.py
```diff
@@ -0,0 +1,27 @@
+PRIMITIVES = set([
```
scottbelden: What do you think about changing this file name to `_schema_common.py`? I think that might better describe that this file is common to both the pyx and py schema files.

scottbelden:
> Right. It's potentially confusing that these other modules do not begin with underscores, even though they are conceptually private. However, I don't want to change more than necessary, so I'm not proposing we address that now (or maybe ever).

Agreed. It should probably be cleaned up at some point. I'll make an issue and hopefully get to it sometime.

fastavro/_reader.pyx
barrywhart: Sounds reasonable. I will do it. I think technically things are probably okay as is, but at best it feels pretty odd. I had even considered whether it was possible to only install one set of files or the other. That's probably possible but pretty non-standard.

fastavro/_schema_public.py
barrywhart: That sounds like a better name. I'll change it.

barrywhart:
>Thanks for all the patience.

No problem. It's a complex set of changes, I am new to the code, and lots of people use this library so we don't want to mess anything up. I'm sure I have put people through harder reviews. :-)

fastavro/_schema_public.py
barrywhart: Done.

fastavro/_reader.pyx
barrywhart: I'm looking into this. In trying this, I'm running into a conflicting symbols issue again, where `fastavro.reader` is a module when the code wants a function. I think I had a solution for this earlier; I'll see if I can dig that up. Wacky interactions...

barrywhart: I found a solution (or maybe just a workaround) for this. You'll see it in my next commit...

barrywhart: Ironically, the workaround requires importing the regular `.py` files, not in order to use anything from them, but to prevent them from being imported later by other code in a problematic way. :-|

barrywhart:
Okay, @scottbelden, I think I have addressed your two most recent comments.

kkirsanov:
Not looking the code checked on my production data (15+ schemes) - everything is ok

scottbelden:
The dummy imports in the `__init__.py` can be fixed by https://github.com/tebeka/fastavro/issues/115 so I'm going to squash and merge this. Thanks!

barrywhart:
Great, thanks! I hope to look into similar changes for the reader at some point. The actual Cython work should be pretty straightforward now that we've worked out the structural and packaging issues.

Issue #115, opened 2017-11-22
Refactor reader/writer/etc to import from the python or cython implementation


scottbelden:
After https://github.com/tebeka/fastavro/pull/114 the structure will essentially be something like `fastavro/reader.py` (python implementation) and `fastavro/_reader.pyx` (cython implementation) and `fastavro/__init__.py` does import magic like so:

```python
try:
    from . import _reader
except ImportError as e:
    from . import reader as _reader

schemaless_reader = _reader.schemaless_reader
```

This means that if you do `from fastavro import schemaless_reader` you will cython implementation and fall back to python if it can't import the cython one.

However, there are probably codebases that do things like `from fastavro.reader import schemaless_reader`. In that case it will always get the python implementation.

One might argue that people should be importing from fastavro to let `__init__.py` figure out the right one, but I think `from fastavro.reader import schemaless_reader` is a natural thing to do and so we should help people in those cases.

I think the way to do this is to change `fastavro/reader.py` to `fastavro/_reader.py` and then have `fastavro/reader.py` do a similar thing as `__init__.py` (i.e. try/except the import and expose the public API). At that point the `try/except` could probably be removed from `__init__.py` and it could just `import reader` as `reader` would pick the right implementation.

PR #117, opened 2017-11-25
rework imports

scottbelden:
Resolves https://github.com/tebeka/fastavro/issues/115


scottbelden:
This changes the directory structure so that `fastavro/reader.py`, `fastavro/writer.py`, and `fastavro/schema.py` are wrapper modules that handle importing from either the cython or python implementations. The means that if someone import from `fastavro.reader` or others then they will get the correct implementation (whereas before they would always get the python implementation).

There are also some `_X_common.py` files that have variables that are common to both the cython and python implementations to reduce duplication.

tebeka:
@scottbelden Can you email me at miki.tebeka@gmail.com?

PR #118, opened 2017-11-26
Cython-based implementation of reader module (~25% faster)

barrywhart:
This PR speeds up `_reader.pyx` by applying similar Cython techniques as those seen in `_writer.pyx`. There is a definite speed improvement (about 25%), although less dramatic than for the writer.

Here are some benchmark numbers using the 10,000-record cases from the `benchmark.py` script @scottbelden had provided.

```
Previous release:

==

This PR:
... 50 runs averaged 0.00834895133972 seconds
... 50 runs averaged 0.0154855728149 seconds

... 50 runs averaged 0.0528964567184 seconds
... 50 runs averaged 0.0909136104584 seconds
```


scottbelden:
I merged https://github.com/tebeka/fastavro/pull/117 which caused a merge conflict here.

barrywhart:
I'll fix it.

:-)

IStooges

barrywhart:
The merge conflicts have been addressed. This is ready for review.

barrywhart:
Fixed another merge conflict.

scottbelden:
I'll try to give this a look over probably tomorrow. At a glance it looks good. I think one initial comment was that perhaps we should move `ReadError` from the common file to the `_reader.pyx` as I think it's only used in that file.

barrywhart:
Good point about `ReadError` -- I'll change that.

I was thinking, maybe we should update the README with the latest Cython performance numbers. Do you know how the current numbers were generated?

barrywhart:
I moved the `ReadError` class.

scottbelden:
I'm not sure how the performance numbers were generated. @tebeka would have more of an idea but I'm assuming whatever script was used might not exist anymore. I did a quick modification of my benchmark script to time it against the standard avro implementation and with the cython code (master, not this PR) I was seeing writes at 14 times faster and reads at 8 times faster. With pure python I see write and reads about 2-3 times faster that the standard avro. I will probably commit this script and the other benchmark script at some point so there is something easily accessible in the future, but until then here is what I have: https://gist.github.com/scottbelden/4556268d0ac5ed39025d29ae0cc715b8

barrywhart:
Great, I'll try and run that script. If I don't get around to it, it's not a big deal to me; I'm fine with merging without changing the readme.

barrywhart:
The build failure on Python 3.4 may be an occurrence of this [Cython bug](https://github.com/cython/cython/issues/1894). I'll try the C macro workaround suggested thereto see if it helps.

barrywhart:
I think I have resolved the Cython issue with Python 3.4. I downgraded to Cython 0.25 as suggested [here](https://github.com/openai/mujoco-py/issues/131#issuecomment-332636429). Hopefully at some point they'll get this working. The specific problem is a bit beyond me unless I take a deep dive into the generated C code.

Given that:
* Python 3.5 was released over 2 years ago
* This Cython issue does not arise in any other Python version (2.7, 3.5, 3.6)

it may be worth considering dropping support for Python 3.4.

scottbelden:
Is this only happening because of the move of the `ReadError` class? If so maybe it's better to keep it in the common folder so that we can use the more recent version of cython.

As for dropping support for 3.4, I wouldn't have any qualms, but 3.4 is still supported until 2019 (https://docs.python.org/devguide/#status-of-python-branches) so unless it causes major headaches we should probably keep it. I'm not sure how to get statistics on what versions are downloaded, but if we did that and found that 3.4 isn't downloaded much then that could be an argument for dropping support as well.

barrywhart:
I don't think it has anything to do with moving `ReadError`. I'm seeing the problem with the new `ReaderBase` class, *probably* (I'm guessing) because it is the first "native" (i.e. `cdef`) class in the file.

I did some tinkering to try and work around the problem with C preprocessor settings and/or a couple of "wild guess" tweaks to `_reader.pyx`. None of these helped, but downgrading did.

Here are a few possibilities to consider:
* `fastavro` requires Cython 0.25 at installation time.
  * Pros: Simple, works with all Python versions
  * Cons: Higher potential for Cython version incompatibilities with other libraries
* On Python 3.4, `fastavro` requires Cython 0.25 at installation time. Other Python versions use any version (0.27.3 or later)
  * Pros: Probably reduces risk of Cython version incompatibilities since we're only requiring the older version on Python 3.4
  * Cons: Same as above, but with lower risk. `setup.py` is a little more complex, as it has to check the Python interpreter version.
* Do not distribute the `.pyx` files, only distribute the generated `.c` files
  * Pros: This is the recommended way of distributing Cython extensions. Because Cython does not have to be installed in the target environment, there is no risk of Cython version issues arising (either with the generated C code being bad or with different libraries wanting different Cython versions installed)
  * Cons: Will require tweaking `setup.py` and possibly `publish.sh` to generate the `.c` files and omit the `.pyx` files from the source package. I haven't done this before, so it'd require a bit of tinkering, but probably no more than a couple hours work.

barrywhart:
BTW, I computed PyPI download stats for `fastavro`. Google makes this information available on BigQuery. There are a few hundred 3.4 downloads in the past month. Pretty light, but there are some.

```
file_project,details_python,downloads
fastavro,2.7.12,12754
fastavro,,12286
fastavro,2.7.11,2918
fastavro,2.7.10,2660
fastavro,3.5.2,1153
fastavro,3.5.4,1131
fastavro,3.6.2,1079
fastavro,2.7.6,932
fastavro,3.6.3,671
fastavro,2.7.13,590
fastavro,3.4.3,381
fastavro,3.6.1,375
fastavro,2.7.5,282
fastavro,2.7,253
fastavro,2.7.14,224
fastavro,2.7.9,214
fastavro,3.5.3,119
fastavro,3.5.1,81
fastavro,3.4.2,77
fastavro,3.6.0,75
fastavro,3.6,48
fastavro,2.7.12+,23
fastavro,3.4.5,20
fastavro,2.6.6,19
fastavro,3.4.0,11
fastavro,3.5,9
fastavro,3.5.0,8
fastavro,3.4.1,6
fastavro,2.7.8,4
fastavro,3.5.4rc1,4
fastavro,3.6.0rc1,3
fastavro,3.4.7,2
fastavro,2.7.0,1
fastavro,3.3.5,1
fastavro,2.6,1
fastavro,2.7.3,1
fastavro,2.7.1,1
fastavro,3.5.2+,1
fastavro,3.4.4,1
```

This is the query I ran:
```
SELECT
  file.project,
  details.python,
  COUNT(*) downloads
FROM
  TABLE_DATE_RANGE(
    [the-psf:pypi.downloads],
    DATE_ADD(CURRENT_TIMESTAMP(), -31, "day"),
    DATE_ADD(CURRENT_TIMESTAMP(), -1, "day")
  )
WHERE
  file.project = 'fastavro'
GROUP BY file.project, details.python
ORDER BY downloads desc
LIMIT 1000
```

fastavro/_reader.pyx
```diff
@@ -18,14 +18,14 @@ except ImportError:
...
-    MemoryIO, xrange, btou, iteritems, is_str, str2ints, fstint
+    btou, iteritems, is_str, str2ints, fstint
 )
 from ._schema import (
     extract_record_type, acquaint_schema, populate_schema_defs,
     extract_logical_type
 )
 from ._reader_common import (
-    SchemaResolutionError, MAGIC, SYNC_SIZE, HEADER_SCHEMA
+    ReadError, SchemaResolutionError, MAGIC, SYNC_SIZE, HEADER_SCHEMA
```
scottbelden: How is this not throwing an `ImportError`? Wasn't it removed from `_reader_common`?

fastavro/_reader.pyx
```diff
@@ -18,14 +18,14 @@ except ImportError:
     import json
 
 from ._six import (
-    MemoryIO, xrange, btou, iteritems, is_str, str2ints, fstint
```
scottbelden: What's the reason for not using the `xrange` imported from `_six`?

barrywhart: Cython has special-case code for loops that use `range()` rather than `xrange()`, so they run much faster. You can see some info in the Cython docs [here](http://docs.cython.org/en/latest/src/reference/language_basics.html#for-loops), although they are not that explicit about the fact you only get the faster speed by using `range()` (not `xrange()`).

I think 70-80% of the speed improvement in this PR comes from this one simple change.

fastavro/_reader.pyx
barrywhart: Ah, this was tricky! It actually *was* throwing an `ImportError`, but in `reader.py`, we catch `ImportError` and fall back to the pure Python implementation, which masks the error. This is nasty. I am adding an automated test to catch this: it will verify that in CPython, we are loading the Cython module, and in Pypy, we are loading the `.py` version.

```

```

barrywhart:
I think the "hidden" import error bug is fixed. The build is running slowly, but has passed on Python 3.6 and both PyPy versions. Fingers crossed...

fastavro/_reader.pyx
scottbelden: Ah, that explains it. And thanks for the test.

fastavro/_reader.pyx
scottbelden: Ah, okay. In that case we might want to change `_six.pyx` to do this for us (i.e. set `xrange = range` for both py2 and py3). Right now nothing else imports `xrange` from `six` but if that were to happen in the future in the `schema` or `writer` I could imagine someone might try to use the `xrange` from `_six` and not know about these speedups.

However, I don't think that's worth holding this PR up.

scottbelden:
Thanks again! Merging this and then will try to get a release out shortly.

fastavro/_reader.pyx
barrywhart: I looked into doing that. Although it logically makes sense that would be okay, it doesn't work as intended. In Cython, `for var in range()` is a special pattern Cython looks for and translates to a native C loop. With `xrange` mapped to the underlying `range` function, I think Cython does more of a literal translation instead of substituting the more efficient pattern. It's possible there's a way to make the alias do what we'd want, but I don't know of one.

You can see the difference here in the generated C code:

Using `range()` directly:

Cython:
```
        for offset in range(size):
```

Resulting C code:
```
    __pyx_t_6 = __Pyx_PyInt_As_long(__pyx_v_size); if (unlikely((__pyx_t_6 == (long)-1) && PyErr_Occurred())) __PYX_ERR(0, 197, __pyx_L1_error)
    for (__pyx_t_7 = 0; __pyx_t_7 < __pyx_t_6; __pyx_t_7+=1) {
      __pyx_v_offset = __pyx_t_7;
```

Calling `range()` via a function that aliases to `range()`:

Cython:
```
         for offset in xrange(size):
```

Resulting C code:
```
    __pyx_t_2 = __Pyx_GetModuleGlobalName(__pyx_n_s_xrange); if (unlikely(!__pyx_t_2)) __PYX_ERR(0, 197, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_2);
    __pyx_t_1 = NULL;
    if (CYTHON_UNPACK_METHODS && unlikely(PyMethod_Check(__pyx_t_2))) {
      __pyx_t_1 = PyMethod_GET_SELF(__pyx_t_2);
      if (likely(__pyx_t_1)) {
        PyObject* function = PyMethod_GET_FUNCTION(__pyx_t_2);
        __Pyx_INCREF(__pyx_t_1);
        __Pyx_INCREF(function);
        __Pyx_DECREF_SET(__pyx_t_2, function);
      }
    }
    if (!__pyx_t_1) {
      __pyx_t_3 = __Pyx_PyObject_CallOneArg(__pyx_t_2, __pyx_v_size); if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 197, __pyx_L1_error)
      __Pyx_GOTREF(__pyx_t_3);
    } else {
      #if CYTHON_FAST_PYCALL
      if (PyFunction_Check(__pyx_t_2)) {
        PyObject *__pyx_temp[2] = {__pyx_t_1, __pyx_v_size};
        __pyx_t_3 = __Pyx_PyFunction_FastCall(__pyx_t_2, __pyx_temp+1-1, 1+1); if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 197, __pyx_L1_error)
        __Pyx_XDECREF(__pyx_t_1); __pyx_t_1 = 0;
        __Pyx_GOTREF(__pyx_t_3);
      } else
      #endif
      #if CYTHON_FAST_PYCCALL
      if (__Pyx_PyFastCFunction_Check(__pyx_t_2)) {
        PyObject *__pyx_temp[2] = {__pyx_t_1, __pyx_v_size};
        __pyx_t_3 = __Pyx_PyCFunction_FastCall(__pyx_t_2, __pyx_temp+1-1, 1+1); if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 197, __pyx_L1_error)
        __Pyx_XDECREF(__pyx_t_1); __pyx_t_1 = 0;
        __Pyx_GOTREF(__pyx_t_3);
      } else
      #endif
      {
        __pyx_t_4 = PyTuple_New(1+1); if (unlikely(!__pyx_t_4)) __PYX_ERR(0, 197, __pyx_L1_error)
        __Pyx_GOTREF(__pyx_t_4);
        __Pyx_GIVEREF(__pyx_t_1); PyTuple_SET_ITEM(__pyx_t_4, 0, __pyx_t_1); __pyx_t_1 = NULL;
        __Pyx_INCREF(__pyx_v_size);
        __Pyx_GIVEREF(__pyx_v_size);
        PyTuple_SET_ITEM(__pyx_t_4, 0+1, __pyx_v_size);
        __pyx_t_3 = __Pyx_PyObject_Call(__pyx_t_2, __pyx_t_4, NULL); if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 197, __pyx_L1_error)
        __Pyx_GOTREF(__pyx_t_3);
        __Pyx_DECREF(__pyx_t_4); __pyx_t_4 = 0;
      }
    }
    __Pyx_DECREF(__pyx_t_2); __pyx_t_2 = 0;
    if (likely(PyList_CheckExact(__pyx_t_3)) || PyTuple_CheckExact(__pyx_t_3)) {
      __pyx_t_2 = __pyx_t_3; __Pyx_INCREF(__pyx_t_2); __pyx_t_6 = 0;
      __pyx_t_7 = NULL;
    } else {
      __pyx_t_6 = -1; __pyx_t_2 = PyObject_GetIter(__pyx_t_3); if (unlikely(!__pyx_t_2)) __PYX_ERR(0, 197, __pyx_L1_error)
      __Pyx_GOTREF(__pyx_t_2);
      __pyx_t_7 = Py_TYPE(__pyx_t_2)->tp_iternext; if (unlikely(!__pyx_t_7)) __PYX_ERR(0, 197, __pyx_L1_error)
    }
    __Pyx_DECREF(__pyx_t_3); __pyx_t_3 = 0;
    for (;;) {
      if (likely(!__pyx_t_7)) {
        if (likely(PyList_CheckExact(__pyx_t_2))) {
          if (__pyx_t_6 >= PyList_GET_SIZE(__pyx_t_2)) break;
          #if CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS
          __pyx_t_3 = PyList_GET_ITEM(__pyx_t_2, __pyx_t_6); __Pyx_INCREF(__pyx_t_3); __pyx_t_6++; if (unlikely(0 < 0)) __PYX_ERR(0, 197, __pyx_L1_error)
          #else
          __pyx_t_3 = PySequence_ITEM(__pyx_t_2, __pyx_t_6); __pyx_t_6++; if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 197, __pyx_L1_error)
          __Pyx_GOTREF(__pyx_t_3);
          #endif
        } else {
          if (__pyx_t_6 >= PyTuple_GET_SIZE(__pyx_t_2)) break;
          #if CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS
          __pyx_t_3 = PyTuple_GET_ITEM(__pyx_t_2, __pyx_t_6); __Pyx_INCREF(__pyx_t_3); __pyx_t_6++; if (unlikely(0 < 0)) __PYX_ERR(0, 197, __pyx_L1_error)
          #else
          __pyx_t_3 = PySequence_ITEM(__pyx_t_2, __pyx_t_6); __pyx_t_6++; if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 197, __pyx_L1_error)
          __Pyx_GOTREF(__pyx_t_3);
          #endif
        }
      } else {
        __pyx_t_3 = __pyx_t_7(__pyx_t_2);
        if (unlikely(!__pyx_t_3)) {
          PyObject* exc_type = PyErr_Occurred();
          if (exc_type) {
            if (likely(__Pyx_PyErr_GivenExceptionMatches(exc_type, PyExc_StopIteration))) PyErr_Clear();
            else __PYX_ERR(0, 197, __pyx_L1_error)
          }
          break;
        }
        __Pyx_GOTREF(__pyx_t_3);
      }
      __pyx_t_8 = __Pyx_PyInt_As_int(__pyx_t_3); if (unlikely((__pyx_t_8 == (int)-1) && PyErr_Occurred())) __PYX_ERR(0, 197, __pyx_L1_error)
      __Pyx_DECREF(__pyx_t_3); __pyx_t_3 = 0;
      __pyx_v_offset = __pyx_t_8;
/* … */
    }
    __Pyx_DECREF(__pyx_t_2); __pyx_t_2 = 0;
```

scottbelden: That's a shame, but good to know. Thanks.

PR #120, opened 2017-11-27
add ujson as an extras_require

scottbelden:
@barrywhart Is this what you were referring to with the `extras_require`?


barrywhart:
Yes, exactly! I wouldn't have included so much detail when I mentioned it if I realized fastavro was already using the same technique for snappy compression.

barrywhart:
:shipit:

Issue #126, opened 2017-12-01
Add testing for Windows


scottbelden:
Travis doesn't support this yet, but it looks like [AppVeyor](https://www.appveyor.com/) might work.

Issue #125, opened 2017-11-30
0.16.3 has issues when parsing double or np.float64 on Windows platforms


An-Li-GitHub:
Hey guys,

i did not have this issue while using 0.16.1 nor 0.14.11.

code to replicate

envrionment: python 2.7.13
cython: 0.25.2
fastavro: 0.16.3

============================================================
import fastavro as avro
from fastavro import writer

schema = {
    'doc': 'A weather reading.',
    'name': 'Weather',
    'namespace': 'test',
    'type': 'record',
    'fields': [
        {'name': 'station', 'type': 'string'},
        {'name': 'time', 'type': 'long'},
        {'name': 'temp', 'type': 'int'},
        {'name': 'test_float', 'type': 'double'}
    ],
}

with open('weather_out.avro', 'wb') as out:
    writer(out, schema, [{u'station': u'011990-99999',
  u'temp': 0L,
  'test_float': 0.21334215134123513,
  u'time': -714214260L},
 {u'station': u'011990-99999',
  u'temp': 22L,
  'test_float': 0.21334215134123513,
  u'time': -714213259L},
 {u'station': u'011990-99999',
  u'temp': -11L,
  'test_float': 0.21334215134123513,
  u'time': -714210269L},
 {u'station': u'012650-99999',
  u'temp': 111L,
  'test_float': 0.21334215134123513,
  u'time': -714208170L}])

with open('weather_out.avro', 'rb') as fo:
    reader = avro.reader(fo)
    schema = reader.schema
    
    for x in reader:
        print x

scottbelden:
Hmm, do you mind expanding on what is wrong? I ran your code and it appeared to work. It looks like the output matches the input (unless I didn't look close enough). This is what I got...

```python
In [5]: %paste
import fastavro as avro
from fastavro import writer

schema = {
'doc': 'A weather reading.',
'name': 'Weather',
'namespace': 'test',
'type': 'record',
'fields': [
{'name': 'station', 'type': 'string'},
{'name': 'time', 'type': 'long'},
{'name': 'temp', 'type': 'int'},
{'name': 'test_float', 'type': 'double'}
],
}

with open('weather_out.avro', 'wb') as out:
    writer(out, schema, [{u'station': u'011990-99999',
    u'temp': 0L,
    'test_float': 0.21334215134123513,
    u'time': -714214260L},
    {u'station': u'011990-99999',
    u'temp': 22L,
    'test_float': 0.21334215134123513,
[...]

PR #127, opened 2017-12-02
add appveyor testing for windows

scottbelden:
Resolves https://github.com/tebeka/fastavro/issues/125
Resolves https://github.com/tebeka/fastavro/issues/126

@barrywhart I needed to change some of the `long`s to `long long` as it seems that for windows a `long` is only 4 bytes whereas on linux it is 8 bytes (which was causing some tests to fail on windows). I wasn't sure if this was the best approach and thought you might have more experience here.

### DIFF
--- a/tests/test_logical_types.py
+++ b/tests/test_logical_types.py
@@ -6,6 +6,8 @@ from decimal import Decimal
 from io import BytesIO
 from uuid import uuid4
 import datetime
+import sys
+import os
 
 
 schema = {
@@ -76,6 +78,8 @@ def test_logical_types():
             == data2['time-millis'].microsecond)
 
 
+@pytest.mark.skipif(os.name == 'nt' and sys.version_info[:2] == (3, 6),
+                    reason='Python Bug: https://bugs.python.org/issue29097')
 def test_not_logical_ints():
     data1 = {
         'date': 1,
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,22 @@
+branches:
+    only:
+        - master
+
+environment:
+    matrix:
+        - PYTHON: "C:\\Python27-x64"
+        # Can't build on 3.4 due to https://github.com/appveyor/ci/issues/1699
+        # and/or https://github.com/appveyor/ci/issues/968
+        # - PYTHON: "C:\\Python34-x64"
+        - PYTHON: "C:\\Python35-x64"
+        - PYTHON: "C:\\Python36-x64"
+
+install:
+    - "%PYTHON%\\python.exe -m pip install pytest flake8 check-manifest Cython"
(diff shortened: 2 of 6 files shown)


tests/test_logical_types.py
```diff
@@ -76,20 +78,25 @@ def test_logical_types():
...
-def test_not_logical_ints():
-    data1 = {
-        'date': 1,
-        'timestamp-millis': 2,
-        'timestamp-micros': 3,
-        'uuid': uuid4(),
-        'time-millis': 4,
-        'time-micros': 5,
+if os.name == 'nt' and sys.version_info[:2] == (3, 6):
+    @pytest.mark.skip(reason='Python Bug: https://bugs.python.org/issue29097')
```
barrywhart: You may want to look into using `skipif`.

https://docs.pytest.org/en/latest/skipping.html

barrywhart:
I gave it a quick look. It seems reasonable. So the "long long" type works okay on Linux, too?

If so, its fine. If not, then in regular C or C++, I would typically use a typedef to encapsulate types that may differ by platform. That way those details are isolated to one small part of the code. I'm not familiar with how that works in Cython, but it is probably possible. The idea would be to use types that specify the size, e.g. long64.

barrywhart:
More detail about the above... the `[ctypedef](http://cython.readthedocs.io/en/latest/src/userguide/language_basics.html)` keyword lets you define aliases for types. I think by combining `IF` statements (see the example in the same page linked above), we could set up aliases for 32 and 64-bit signed and unsigned integer types and use those throughout the `.pyx` files.

scottbelden:
Changed the test to clean up the conditional skip.

The `long long` works as far as I can tell. The original tests continue to pass. I added the relevant information to https://github.com/tebeka/fastavro/issues/128 as potential future work.

barrywhart:
About `long long`, it probably does work fine, since it sounds like `long long` and `long` are the same in Linux. The work I suggested could be helpful in the future if we encounter other OSes/compilers with incompatible types or behavior.

tests/test_logical_types.py
```diff
@@ -77,6 +79,10 @@ def test_logical_types():
 
 
 def test_not_logical_ints():
+    # This test doesn't work on Windows with python 3.6
+    if os.name == 'nt' and sys.version_info[:2] == (3, 6):
+        pytest.skip('Python Bug: https://bugs.python.org/issue29097')
```
barrywhart: If you use `skipif`, you can pass the condition as part of the `skipif` function call. I think what you have works, this is more of a style thing.

barrywhart: https://docs.pytest.org/en/latest/skipping.html#id1

Issue #48, opened 2016-05-18
Validation of Unions


NickG123:
Hey,
I'm currently running into the following problem:
With a schema like this:

```
protocol test {
  record a {
    int x;
    int y;
  }

  record b {
    int x;
    int y;
    union {null, int} z;
  }

  record test {
    union {a, b} val;
  } 
}
```

I can create a file with two objects:

```
>>> data = StringIO.StringIO()
>>> fastavro.writer(data, schema, [{'val': {'x': 1, 'y': 2}},{'val': {'x': 3, 'y':4, 'z':5}}])
```

However, because of the way unions are validated, both 'val's are serialized as type 'a', and the 'z' field of the second object is silently dropped.

This can be confirmed by reading in the objects:

```
>>> data.seek(0)
>>> reader = fastavro.reader(data)
>>> next(reader)
{u'val': {u'y': 2, u'x': 1}}
>>> next(reader)
{u'val': {u'y': 4, u'x': 3}}
```

Do you have any suggestions for a way of dealing with this problem?

Thanks,
Nick

tebeka:
Thanks for reporting. I'll try to have a look this weekend.

scottbelden:
Someone posted a similar problem with the standard avro library here: https://issues.apache.org/jira/browse/AVRO-1777

Their suggestion was to keep track of the number of matching fields and in the end go with the one that matched the most. It seems pretty reasonable. Alternatively, one could go with a much more strict approach and force the keys in the dictionary to exactly match the fields in the schema, but that would break people that are making the use of the fact that extra data in the dictionary is ignored.

tebeka:
This should be fixed in 0.9.10

movermeyer:
This was not fixed. Tested with fastavro-0.9.11 using the test case from this issue.

tebeka:
Hmm. OK, will check again.

PR #129, opened 2017-12-07
Fix union schema resolution in writer

Artimi:
This fixes https://github.com/tebeka/fastavro/issues/48 as @scottbelden suggested in https://github.com/tebeka/fastavro/issues/48#issuecomment-220317143 . I had to change return value of `writer.validate` from `bool` to `int` so it can specify the score of validation, i.e. number of fields that validated against given `schema`. Then in `write_union` the schema is chosen not only if the schema validates but the one with highest number of validated fields is chosen.

### DIFF
--- a/fastavro/_writer.pyx
+++ b/fastavro/_writer.pyx
@@ -437,6 +437,10 @@ cpdef write_union(bytearray fo, datum, schema):
     zero-based position within the union of the schema of its value. The value
     is then encoded per the indicated schema within the union."""
 
+    cdef int best_match_index
+    cdef int most_fields
+    cdef int index
+    cdef int fields
     if isinstance(datum, tuple):
         (name, datum) = datum
         for index, candidate in enumerate(schema):
@@ -449,12 +453,22 @@ cpdef write_union(bytearray fo, datum, schema):
             raise ValueError(msg)
     else:
         pytype = type(datum)
+        best_match_index = -1
+        most_fields = -1
         for index, candidate in enumerate(schema):
             if validate(datum, candidate):
-                break
-        else:
+                if extract_record_type(candidate) == 'record':
+                    fields = len(candidate['fields'])
+                    if fields > most_fields:
+                        best_match_index = index
+                        most_fields = fields
+                else:
+                    best_match_index = index
+                    break
+        if best_match_index < 0:
             msg = '%r (type %s) do not match %s' % (datum, pytype, schema)
             raise ValueError(msg)
+        index = best_match_index
 
     # write data
     write_long(fo, index)
--- a/fastavro/_writer_py.py
+++ b/fastavro/_writer_py.py
@@ -395,12 +395,22 @@ def write_union(fo, datum, schema):
             raise ValueError(msg)
     else:
         pytype = type(datum)
+        best_match_index = -1
+        most_fields = -1
         for index, candidate in enumerate(schema):
             if validate(datum, candidate):
-                break
-        else:
+                if extract_record_type(candidate) == 'record':
+                    fields = len(candidate['fields'])
+                    if fields > most_fields:
+                        best_match_index = index
+                        most_fields = fields
+                else:
+                    best_match_index = index
+                    break
+        if best_match_index < 0:
             msg = '%r (type %s) do not match %s' % (datum, pytype, schema)
             raise ValueError(msg)
+        index = best_match_index
 
     # write data
     write_long(fo, index)
--- a/tests/test_fastavro.py
+++ b/tests/test_fastavro.py
@@ -990,3 +990,53 @@ def test_writer_class_split_files(tmpdir):
         new_interim_record_counts.append(len(new_records))
     assert new_records == records
     assert interim_record_counts == new_interim_record_counts
+
+
+def test_union_records():
+    #
+    schema = {
+        'name': 'test_name',
+        'namespace': 'test',
+        'type': 'record',
+        'fields': [
+            {
+                'name': 'val',
+                'type': [
+                    {
+                        'name': 'a',
+                        'namespace': 'common',
+                        'type': 'record',
+                        'fields': [
+                            {'name': 'x', 'type': 'int'},
+                            {'name': 'y', 'type': 'int'},
+                        ],
+                    },
+                    {
+                        'name': 'b',
+                        'namespace': 'common',
+                        'type': 'record',
+                        'fields': [
+                            {'name': 'x', 'type': 'int'},
+                            {'name': 'y', 'type': 'int'},
+                            {'name': 'z', 'type': ['null', 'int']},
+                        ],
+                    }
+                ]
+            }
+        ]
+    }
+
+    data = [{
+        'val': {
+            'x': 3,
+            'y': 4,
+            'z': 5,
+        }
+    }]
+    new_file = MemoryIO()
+    fastavro.writer(new_file, schema, data)
+    new_file.seek(0)
+
+    new_reader = fastavro.reader(new_file)
+    new_records = list(new_reader)
+    assert new_records == data


scottbelden:
@tebeka Do you know why travis didn't run here?

scottbelden:
Thanks for the PR. I haven't done any profiling, but I think it might be simpler to go with the approach that was mentioned in the AVRO JIRA ticket. Essentially something like this:

```python
cpdef write_union(bytearray fo, datum, schema):
    """A union is encoded by first writing a long value indicating the
    zero-based position within the union of the schema of its value. The value
    is then encoded per the indicated schema within the union."""

    if isinstance(datum, tuple):
        ...
    else:
        pytype = type(datum)
        best_match_index = -1
        most_fields = -1
        for index, candidate in enumerate(schema):
            if validate(datum, candidate):
                if extract_record_type(candidate) == 'record':
                    fields = len(candidate['fields'])
                else:
                    fields = 1
                if fields > most_fields:
                    best_match_index = index
                    most_fields = fields
        if best_match_index < 0:
            msg = '%r (type %s) do not match %s' % (datum, pytype, schema)
            raise ValueError(msg)
        index = best_match_index

    # write data
    write_long(fo, index)
    write_data(fo, datum, schema[index])
```

fastavro/_writer.pyx
```diff
@@ -449,9 +451,9 @@ cpdef write_union(bytearray fo, datum, schema):
             raise ValueError(msg)
     else:
         pytype = type(datum)
-        for index, candidate in enumerate(schema):
-            if validate(datum, candidate):
-                break
+        validation_scores = [validate(datum, candidate) for candidate in schema]
```
barrywhart: Do you think we could choose the highest-scoring option without building and scanning a temporary list? It could be significantly faster that way. Maybe something like this:
```
cdef int validation_score
cdef int best_validation_score = -1
cdef int i, index, best_index, validation_score
for i, candidate in enumerate(schema):
    validation_score = validate(datum, candidate)
    if validation_score > best_validation_score:
        best_validation_score = validation_score
        best_index = i
if best_validation_score > 0:
    index = best_index
else:
    msg = '%r (type %s) do not match %s' % (datum, pytype, schema)
    raise ValueError(msg)
```

Artimi:
@scottbelden Thanks for that tip I don't know why I missed it. It is surely more elegant. I only changed the else branch for non record types. In previous implementation the first candidate that validates was used and I think we should stick with this (also it will be faster when we won't validate everything when we don't have to).

fastavro/_writer.pyx
Artimi: Thanks for the feedback. I changed it in a way https://github.com/tebeka/fastavro/pull/129#issuecomment-350354076 was suggesting and it includes your remark too.

fastavro/_writer.pyx
```diff
@@ -449,12 +449,22 @@ cpdef write_union(bytearray fo, datum, schema):
...
         for index, candidate in enumerate(schema):
             if validate(datum, candidate):
-                break
-        else:
+                if extract_record_type(candidate) == 'record':
+                    fields = len(candidate['fields'])
+                else:
+                    best_match_index = index
+                    break
+                if fields > most_fields:
```
scottbelden: Can you move this `if` clause to be after `fields = len(candidate['fields'])` since it only applies for that case?

fastavro/_writer_py.py
```diff
@@ -395,12 +395,22 @@ def write_union(fo, datum, schema):
...
         for index, candidate in enumerate(schema):
             if validate(datum, candidate):
-                break
-        else:
+                if extract_record_type(candidate) == 'record':
+                    fields = len(candidate['fields'])
+                else:
+                    best_match_index = index
+                    break
+                if fields > most_fields:
```
scottbelden: Can you move this `if` clause to be after `fields = len(candidate['fields'])` since it only applies for that case?

fastavro/_writer.pyx
```diff
@@ -449,12 +449,22 @@ cpdef write_union(bytearray fo, datum, schema):
             raise ValueError(msg)
     else:
         pytype = type(datum)
+        best_match_index = -1
```
scottbelden: @barrywhart Should `best_match_index`, `most_fields`, and `index` be defined as `cdef int`? I noticed that it ends up with more pure c code when running `cython -a`?

scottbelden:
I'm going to close and re-open this to see if it will run on travis...

fastavro/_writer.pyx
barrywhart: I would say yes.  `cython -a` gives a pretty good rough indication of how the code will perform, and more yellow is generally bad.

fastavro/_writer.pyx
```diff
@@ -449,12 +449,22 @@ cpdef write_union(bytearray fo, datum, schema):
...
+        most_fields = -1
         for index, candidate in enumerate(schema):
             if validate(datum, candidate):
-                break
-        else:
+                if extract_record_type(candidate) == 'record':
+                    fields = len(candidate['fields'])
```
barrywhart: `fields` should also be declared using `cdef int`. This will speed up the `fields > most_fields` comparison later.

tests/test_fastavro.py
```diff
@@ -817,3 +817,52 @@ def test_cython_python():
...
+    }
+
+    data = {
+        'val': {
+            'x': 3,
+            'y': 4,
+            'z': 5,
+        }
+    }
+    encoded = bytearray()
```
barrywhart: I think this will fail in `pypy`, because the pure Python implementation of `dump()` takes a file object rather than a `bytearray` parameter.

I introduced this inconsistency when implementing the Cython writer -- I did not realize that `dump()` is a public function of the library.

We / I need to think about how to address this.

barrywhart: I think the simplest way to fix the library issue would be for `fastavro.dump()` to *always* call the pure Python version `fastavro._writer_py`, since that function already writes to a file object. It will be slower than the Cython version, which may not matter much, depending on how heavily the function is used.

A more proper fix would be for the Cython code to provide two functions, one writing to a file object and the other to a `bytearray`, and have `fastavro.dump()` use the former.

@scottbelden, @tebeka: Can you confirm that `fastavro.dump()` is intended to be used by application code to write single records to file objects? I'd like your thoughts on the above as well. I can create a PR for this if you want; I think it's a pretty simple change (< 1 hour).

scottbelden:
@Artimi Can you change the name of the branch you are using so that it doesn't have a `/` in it. I think that is causing travis to not run the tests. When I pushed a branch to get the tests to run it failed on pypy as @barrywhart expected it might. I think for this PR it would be best to change the test to use the `writer` and `reader` like most of the other tests in the file. As for the issue with `dump` and `load` being different with pure python vs cython I'll make an issue and try to give some thoughts later.

tests/test_fastavro.py
barrywhart: I have a PR up which should address this: https://github.com/tebeka/fastavro/pull/133.

Once it is merged, the `bytearray` in this code should be replaced with a file object. With that change, it should now work in both CPython and Pypy.

fastavro/_writer.pyx
Artimi: I've changed `best_match_index`, `index`, `most_fields` and `fields` to be declared by `cdef int` as suggested.

fastavro/_writer.pyx
Artimi: Done, thanks for suggestion.

Artimi:
@scottbelden It seems that travis is now running even with `/` in branch name. I would rather not rename it because AFAIK I would need to create another PR and we would lost this conversation.
I've changed the test so it is in the same manner as others in that class and now it seems passing even for travis. Thanks for dealing with that `dump` issue.

scottbelden:
Strange... I'm not sure why travis didn't run initially and then started running. Regardless, thanks for the PR!

scottbelden:
FYI, I haven't released it yet but will do so shortly.

Issue #130, opened 2017-12-07
Provide a documented way to limit maximum file size when writing


barrywhart:
TL;DR
* Make `Writer` class public?
* Provide a way for users of`Writer` to know each time a block is written?

We often dump very large data sets to Avro. We need a generic way to limit the output file size to something reasonable for Hadoop (say, 64-256MB). This is tricky since records can be of any size.

We found that we could use `fastavro`'s `Writer` class in the following way:

```
        writer.write(record)

        # Block count of zero indicates fastavro just reached an internal max
        # size and ran Writer.dump(). This is a good time to consider starting
        # a new file.
        if writer.block_count == 0 and output_file.tell() >= MAX_FILE_SIZE:
            output_file.close()
            writer = Writer(...)
```

This worked okay but has a couple of issues:
* The `Writer`class is not part of the public interface of the library.
* It relies on internal knowledge of the `Writer` class:
  * `block_count` and `io` fields
  * `dump()` function

Also (irony), my [recent PR](https://github.com/tebeka/fastavro/pull/114) which added `_writer.pyx` breaks the above approach because the fields are now private (due to being declared `cdef` rather than `cpdef`.

I would like to restore the above capability and am happy to contribute a PR. I'd like to hear others' thoughts on how to do this. The simplest approach would be to expose those fields to Python, but it doesn't seem like a great design. Another possible approach would update `Writer` to notify when it writes a block, perhaps by adding a new optional `on_write_block_fn` parameter to `Writer`'s constructor.

@scottbelden

scottbelden:
Hmm... I'd be worried about exposing `Writer` for a couple of reasons.

- As you mentioned, it requires knowledge of some internals like `block_count`
- The name is basically the same as the current `writer` which could be confusing. However, the solution to that is probably to change `writer` to `write`

Having `Writer` be public doesn't seem like it would be too bad, but it might require extra thought on what parts of it should be public. For example, I don't know if `dump` should be public as it seems like `flush` should be sufficient.

One idea (not sure if I think it's good or not...) could be to have `writer` keep track of how many bytes it has written and take an option `cutoff` argument to know when it should stop.

The `writer` function would probably need to return the number of records written so that if it was given a list of records the user would know how many records still need to be written. For example:

```python
def writer(..., cutoff=None):
    output = Writer(
        fo,
        schema,
        codec,
        sync_interval,
        metadata,
        validator,
    )

    bytes_written = 0
    records_writen = 0
    for record in records:
[...]

PR #131, opened 2017-12-07
For issue #130: Expose Writer class, add "dirty" property

barrywhart:
Can I get some preliminary feedback on this PR? It implements the first idea discussed in #130.

- [x] Expose `Writer` as a public class
- [x] Make `Writer`'s `.io` field expose a similar API in regular Python and Cython
- [x] Rerun writer performance benchmark to ensure no performance regressions
- [x] Add new tests for `Writer` now that it's public? (Would appreciate some suggestions on this.)

### DIFF
--- a/fastavro/_writer.pyx
+++ b/fastavro/_writer.pyx
@@ -5,7 +5,7 @@
 # Apache 2.0 license (http://www.apache.org/licenses/LICENSE-2.0)
 
 from ._six import utob, long, is_str, iterkeys, itervalues, iteritems, mk_bits
-from .reader import HEADER_SCHEMA, SYNC_SIZE, MAGIC
+from ._reader import HEADER_SCHEMA, SYNC_SIZE, MAGIC
 from ._schema import (
     extract_named_schemas_into_repo, extract_record_type,
     extract_logical_type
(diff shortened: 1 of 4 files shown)


barrywhart:
This PR:

```
... 50 runs averaged 0.00858006954193 seconds
... 50 runs averaged 0.0159693431854 seconds

... 50 runs averaged 0.0575728607178 seconds
... 50 runs averaged 0.0941914415359 seconds
```

Benchmark run from the original `_writer.pyx` PR:
```

```

So the write performance still looks good.

barrywhart:
I made a few changes based on the discussion so far:
* In `Writer`
  * Make most fields private (only in Cython version; Python doesn't support this)
  * Rename `dump()` as `_dump()` to imply it's private
  * Add a `dirty` property so it's possible to implement a cutoff without accessing `block_count` and `io` fields

With these changes, the file size cutoff can now be implemented externally as:
```
        writer.write(record)

        # Block count of zero indicates fastavro just reached an internal max
        # size and ran Writer.dump(). This is a good time to consider starting
        # a new file.
        if not writer.dirty and output_file.tell() >= MAX_FILE_SIZE:
                    output_file.close()
                    writer = Writer(...)
```

(I noticed the original example in the issue I filed was overly complex as well. I will clean that up for a fair comparison.)

barrywhart:
Ok, I reverted the last 3 commits. I'd appreciate any guidance about what level and kind of testing and documentation would be appropriate for this PR.

Thanks!

scottbelden:
This seems good to me. As for testing, I think something that just shows the basic use case would be sufficient for the moment. Something like:

```python
from fastavro.writer import Writer

writer = Writer(file, schema)
for record in records:
    writer.write(record)
writer.flush()

...
# read
...
assert read_records == records
```

This wouldn't test anything new (as the normal `writer` function basically already tests this) but I think would be useful as a proof of "this is basically how it would work"

barrywhart:
I added 3 tests:
* Basic test that illustrates writing records and flushing
* Test that illustrates a useful scenario for unit testing code that uses `Writer`, where small files are preferable so tests run quickly. This test specifies a `sync_interval` of 0, which forces `Writer` to flush automatically after each record.
* Test that illustrates splitting a set of records across two files based on automatic flush behavior. This test approximates the real-world use case of splitting files when they exceed a target size.

scottbelden:
Wow, nice tests. I'm good for merging. @tebeka do you have any thoughts?

tebeka:
@scottbelden Sorry, way too busy this week. Whatever you decide is good for me.

Issue #132, opened 2017-12-12
Deconflict the dump/load API difference between pure python and cython


scottbelden:
Original notes from @barrywhart [here](https://github.com/tebeka/fastavro/pull/129#discussion_r156236682)

> I think this will fail in pypy, because the pure Python implementation of dump() takes a file object rather than a bytearray parameter.
> 
> I introduced this inconsistency when implementing the Cython writer -- I did not realize that dump() is a public function of the library.
> 
> We / I need to think about how to address this.

and

> I think the simplest way to fix the library issue would be for fastavro.dump() to always call the pure Python version fastavro._writer_py, since that function already writes to a file object. It will be slower than the Cython version, which may not matter much, depending on how heavily the function is used.
> 
> A more proper fix would be for the Cython code to provide two functions, one writing to a file object and the other to a bytearray, and have fastavro.dump() use the former.
> 
> @scottbelden, @tebeka: Can you confirm that fastavro.dump() is intended to be used by application code to write single records to file objects? I'd like your thoughts on the above as well. I can create a PR for this if you want; I think it's a pretty simple change (< 1 hour).

I haven't thought about this yet. I'll try to give some thoughts later.

barrywhart:
I will create a PR for this, hopefully in the next day or two. I plan to go for the "more proper" fix. As @scottbelden had pointed out once before, it seems cleaner to try and avoid *ever* importing both the Cython and non-Cython versions of the same module, because it can easily lead to confusing bugs.

scottbelden:
Resolved via https://github.com/tebeka/fastavro/pull/133

PR #133, opened 2017-12-13
Issue 132: Public Cython dump() function should take file object, not bytearray

barrywhart:
Addresses #132


scottbelden:
Thanks again. I was wondering though if it makes sense to keep `write_data` exposed in `fastavro/writer.py`. It seems like `dump` should be the interface. It could be potentially breaking to remove `write_data` from `writer.py` if someone has done `from fastavro.writer import write_data` but I think we'll eventually need a breaking change to rename some modules/functions/classes, so it might make sense at that time.

Thoughts?

barrywhart:
I agree about `write_data`. It seems odd to expose the same functionality under two different names, and `write_data` definitely feels more like an internal name.

Issue #128, opened 2017-12-04
Maybe use cython typedefs to encapsulate types that may differ by platform


scottbelden:
Right now we use `long long` in some places to ensure we have a 64 bit type because `long` on windows is only 32 bits. This seems to be okay at the moment, but we might want to use some cython `typedefs` to do something like the following:

```python
if windows:
    ctypedef unsigned long long U64Long
else:
    ctypedef unsigned long U64Long
```

See https://github.com/tebeka/fastavro/pull/127#issuecomment-348690093 for more information.

scottbelden:
Resolved via https://github.com/tebeka/fastavro/pull/135

PR #135, opened 2017-12-17
Issue 128: Introduce descriptive ctypedefs for various-sized integer types

barrywhart:
Implements #128

### DIFF
--- a/fastavro/_reader.pyx
+++ b/fastavro/_reader.pyx
@@ -56,6 +56,12 @@ AVRO_TYPES = set([
 ])
 
 
+ctypedef int int32
+ctypedef unsigned int uint32
+ctypedef unsigned long long ulong64
+ctypedef long long long64
+
+
 class ReadError(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


fastavro/_reader.pyx
```diff
@@ -274,26 +280,26 @@ cpdef read_double(ReaderBase fo, writer_schema=None, reader_schema=None):
     """
     cdef bytes data
     cdef unsigned char ch_data[8]
-    cdef double_long dl
+    cdef double_ulong64 dl
     data = fo.read(8)
     if len(data) == 8:
         ch_data[:8] = data
         dl.n = (ch_data[0] |
-                (<unsigned long>(ch_data[1]) << 8) |
```
barrywhart: I noticed this code was still using `unsigned long`. This may have been a bug on Windows, where `long` is 32 bits but `double` on Windows is 64 bits. I wonder why tests didn't catch it.

scottbelden: It's possible that we don't have a test that uses the `double` type in such a way that would have exposed the problem.

scottbelden:
Look good. Thanks! I'm going to merge but probably won't do any version release for the moment.

barrywhart:
No problem about delaying the release, this isn't a change I'm relying on. As I mentioned, this *may* fix a bug on Windows, but I'm not 100% sure of that, and our apps run on Linux anyway.

PR #137, opened 2017-12-22
Issue 36: Rename fastavro.writer, other naming cleanup

barrywhart:
PR for #36.

Renaming the reader and writer modules also allows us to simplify some previously tricky import code.


fastavro/_write.pyx
```diff
@@ -30,6 +17,18 @@ from collections import Iterable, Mapping
 from os import urandom, SEEK_SET
 from zlib import compress
 
+from .const import (
```
barrywhart: I moved these imports to respect the recommended PEP-8 import order.

fastavro/_write_py.py
```diff
@@ -35,6 +20,19 @@
 from struct import pack
 from zlib import compress
 
+from .const import (
```
barrywhart: I moved these imports to respect the recommended PEP-8 import order.

scottbelden:
Very much in favor of this. I'm assuming this will break some potential imports that might exist in the wild (maybe something like `from fastavro.reader import reader`)? I'm fine with that but just wanted to know if this was the case for when the inevitable issue is filed.

barrywhart:
>I'm assuming this will break some potential imports that might exist in the wild (maybe something like from fastavro.reader import reader)? I'm fine with that but just wanted to know if this was the case for when the inevitable issue is filed.

Correct, existing code that imports from the `reader` or `writer` modules will break. It's possible there's a way to hack things to avoid this (if it's possible to make a module that acts like a function and vice versa), but I think it's better to just make a clean break with the past here. I don't that many applications will be affected -- seems like most people would have followed the examples in `README.md`.

scottbelden:
Agreed. If anyone gets import errors it should be trivial to fix them. There is a conflict from merging the other PR but once that's resolved we can get this in.

barrywhart:
I fixed the merge conflicts arising from the other PR.

scottbelden:
Thanks again!

barrywhart:
You're welcome! It's a fun project to work on, and very useful for our work!

Issue #33, opened 2015-09-10
Please support JSON encoded AVRO output format


ghost:
https://avro.apache.org/docs/1.7.7/spec.html#json_encoding

tebeka:
Thanks for reporting, I'll take a look.

tebeka:
Sorry for the long delay. How does the current JSON output differ from what's in the spec?

tebeka:
@fcntl Any input on this?

mvallebr:
Could you please consider this example: http://mail-archives.apache.org/mod_mbox/avro-user/201804.mbox/browser

The output differs from what's specified in json_encoding for union types, for instance. 
From https://avro.apache.org/docs/1.7.7/spec.html#json_encoding:

> For example, the union schema ["null","string","Foo"], where Foo is a record name, would encode:
> 
> - null as null;
> - the string "a" as {"string": "a"}; and
> - a Foo instance as {"Foo": {...}}, where {...} indicates the JSON encoding of a Foo instance.

Also, I would expect to be able to write a python dict using `writer` or `schemaless_writer` the same way it's specified in json encoding. This is not the behavior currently.

mvallebr:
> it sounds like you are requesting that they change the specification of JSON encoding

No, not at all, the specification is fine and if followed, it's not ambiguous. What I was requesting them was to change the python library implementation, which doesn't implement JSON encoding specification. 

> Currently, fastavro doesn't implement the JSON encoded avro, but if it did I think it would have to go by what's in the specification.

I couldn't agree more. What I want is the specification implemented.

What can I do to help on this? I have a lot of interest on using an avro library that implements it. I wouldn't like to create a wrapper library on top of neither avro nor fastavro, as I think the base library should support it. 

If you're willing to implement it on a timely manner, I can probably find a way of contributing with tests and test automation. Otherwise, directions on how to contribute with the full feature would also be welcome. I need a library implementing this and I would be very happy if this library is fastavro.

mvallebr:
@scottbelden Did you understand the difference between what we currently have implemented and what is needed?

If you could, please, point me some parts of code I should start looking for to add a failing test case, I could start by contributing to that. With some example test cases failing, I think it should be easier either for you to implement the solution or for me to try to contribute to it.

scottbelden:
[...]

PR #140, opened 2017-12-26
`fastavro` command-line tool crashes if data contains date or date/time values

barrywhart:
This fixes a bug I noticed with the `fastavro` command-line tool dumping an Avro file: If `ujson` is not installed (i.e. if using the built-in `json` module), dumping an Avro file with any of the higher-level logical (Python) types crashes:
* `datetime`
* `date`
* `Decimal`
* `UUID`

This occurs because `json.dumps()` does not support those data types.

I fixed this by scanning each record after it is read but before it is dumped, converting those values to strings. I'm not sure there is a perfect fix here, but this seems reasonable.

Here's an example stack trace of the failure with a `datetime` value:
```
  File "/Users/bhart/.pyenv/versions/fastavro-2.7.13/bin/fastavro", line 11, in <module>
    load_entry_point('fastavro', 'console_scripts', 'fastavro')()
  File "/Users/bhart/dev/fastavro/fastavro/__main__.py", line 54, in main
    json_dump(record, indent)
  File "/Users/bhart/dev/fastavro/fastavro/six.py", line 71, in py2_json_dump
    json.dump(obj, stdout, indent=indent, encoding=_outenc)
  File "/Users/bhart/.pyenv/versions/2.7.13/lib/python2.7/json/__init__.py", line 189, in dump
    for chunk in iterable:
  File "/Users/bhart/.pyenv/versions/2.7.13/lib/python2.7/json/encoder.py", line 434, in _iterencode
    for chunk in _iterencode_dict(o, _current_indent_level):
  File "/Users/bhart/.pyenv/versions/2.7.13/lib/python2.7/json/encoder.py", line 408, in _iterencode_dict
    for chunk in chunks:
  File "/Users/bhart/.pyenv/versions/2.7.13/lib/python2.7/json/encoder.py", line 442, in _iterencode
    o = _default(o)
  File "/Users/bhart/.pyenv/versions/2.7.13/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: datetime.datetime(2017, 12, 24, 23, 56, 23) is not JSON serializable
```

### DIFF
--- a/fastavro/six.py
+++ b/fastavro/six.py
@@ -65,10 +65,13 @@ else:  # Python 2x
     _outenc = getattr(stdout, 'encoding', None) or _encoding
 
     def py2_json_dump(obj, indent):
+        kwargs = {}
+        if indent is not None:
+            kwargs['indent'] = indent
         if _HAS_UJSON:
-            json.dump(obj, stdout, indent=indent)
+            json.dump(obj, stdout, **kwargs)
         else:
-            json.dump(obj, stdout, indent=indent, encoding=_outenc)
+            json.dump(obj, stdout, encoding=_outenc, **kwargs)
 
     def py2_iterkeys(obj):
         return obj.iterkeys()
--- a/fastavro/six.py
+++ b/fastavro/six.py
@@ -29,7 +29,10 @@ if sys.version_info >= (3, 0):
         return bytes(n, encoding)
 
     def py3_json_dump(obj, indent):
-        json.dump(obj, stdout, indent=indent)
+        if _HAS_UJSON:
+            json.dump(obj, stdout, indent=indent)
+        else:
+            json.dump(obj, stdout, indent=indent, encoding=_outenc)
 
     def py3_iterkeys(obj):
         return obj.keys()
--- a/fastavro/__main__.py
+++ b/fastavro/__main__.py
@@ -1,8 +1,23 @@
-import fastavro as avro
-from fastavro.six import json_dump
+import datetime
 from sys import stdout
 
+import fastavro as avro
+from fastavro.six import iteritems, json_dump
+
 encoding = stdout.encoding or "UTF-8"
+
+
+def _clean_json(data):
+    if isinstance(data, dict):
+        for k, v in iteritems(data):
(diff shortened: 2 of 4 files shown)


fastavro/six.py
```diff
@@ -65,10 +65,13 @@ def py2_utob(n, encoding=_encoding):
     _outenc = getattr(stdout, 'encoding', None) or _encoding
 
     def py2_json_dump(obj, indent):
+        kwargs = {}
```
scottbelden: Unrelated to this PR, but I just noticed that `py3_json_dump` doesn't do any checking for `_HAS_UJSON`... it seems like it should, right? If you want to add that to this PR then that'd be great otherwise I'll make an issue and get back to it in the future

scottbelden:
I'm surprised this file code is used. I didn't know anyone used that tool. This seems like a fine solution. I think the real solution would be [to support JSON encoded avro](https://github.com/tebeka/fastavro/issues/33) but that's a much heavier lift.

fastavro/six.py
barrywhart: Will do.

barrywhart:
Yes, I find this tool pretty useful when I want to take a quick glance at some Avro data, e.g. to verify the schema is correct, the data looks reasonable, etc.

About JSON-encoded Avro, can you explain exactly what it is? The spec talks about it, but it jumps immediately into a few specific cases, and I don't get the "big picture". What exactly is it, and what does it look like, especially with logical types?

scottbelden:
I think the specification is pretty bad as far as describing the JSON encoding. At one time I thought about using [this](https://github.com/linkedin/python-avro-json-serializer) as a reference for implementing it here, but I didn't really need the avro-json encoding so I didn't bother. I don't think the standard avro python library implements it either so the only way to really get some examples would be to do it in Java and I didn't feel like doing that.

barrywhart:
Oh, I see -- it's basically JSON but with an extra level of dictionary nesting so the value type can be explicitly encoded as well. That's not something I have encountered a need for, either.

Issue #139, opened 2017-12-26
support for OrderedDict in when writing record types


olibook:
Currently (as 0.16.6), the writer module fails to serialise OrderedDict instances when mapped to a record in a schema when using cpython interpreters.

   ```
avro.schemaless_writer(buff, self.schema, item)
  File "fastavro/_writer.pyx", line 797, in fastavro._writer.schemaless_writer
  File "fastavro/_writer.pyx", line 559, in fastavro._writer.write_data
  File "fastavro/_writer.pyx", line 478, in fastavro._writer.write_record
  File "fastavro/_writer.pyx", line 491, in fastavro._writer.write_record
  File "fastavro/_writer.pyx", line 559, in fastavro._writer.write_data
  File "fastavro/_writer.pyx", line 435, in fastavro._writer.write_union
  File "fastavro/_writer.pyx", line 475, in fastavro._writer.write_union
  File "fastavro/_writer.pyx", line 559, in fastavro._writer.write_data
TypeError: Argument 'datum' has incorrect type (expected dict, got collections.OrderedDict)
```

The use case for supporting ordereddicts as dicts is as follow: we publish data model changes from our django apps to a kafka broker. We prepare the model with django-rest-framework's serialisers to get a "view" of the model. Django-rest-framework returns ordered dicts.

Ordered dicts can be converted to dicts with the dict() constructor. However supporting nested resources requires converting the complete data-structure before sending it to avro's writer.

A trivial test case demonstrating the issue is available here:

https://github.com/olibook/fastavro/commit/b2e8822d70de56e717560ac739b66e640171ca75

it shows that affected versions are python 2.7, 3,4 3.5, 3.6. Unaffected python versions : pypy and pypy3

scottbelden:
@barrywhart It looks like this is due to the cython files defining functions like `write_*` as `cpdef write_record(bytearray fo, dict datum, dict schema)` and so if `datum` is an `OrderedDict` cython thinks it is a different type. I did a quick search seeing if there was a way to have a more generic `mapping` type in cython they would accept all dict-like objects but couldn't find anything right away. Do you have thoughts on how to allow this?

barrywhart:
Interesting. Cython has built-in handling for a handful of native Python types such as `dict`, `list`, `bytes`, and `bytearray`. By this, I mean you can write Python-like code which uses those types directly. Apparently that built-in support requires you to use the exact type, not a subtype.

[...]

PR #141, opened 2017-12-27
Handle records and maps that are OrderedDict rather than dict

barrywhart:
Address #139.

We add a runtime check to `write_record()` and `write_map()`. For `dict`, use faster code. For other types, use equivalent but slower, generic code. There appears to be no meaningful impact on performance.

```
Before:

... 10 runs averaged 0.00915930271149 seconds
... 10 runs averaged 0.0144769668579 seconds

... 10 runs averaged 0.0550956249237 seconds
... 10 runs averaged 0.0886501789093 seconds

After:

... 10 runs averaged 0.00916681289673 seconds
... 10 runs averaged 0.0137780666351 seconds

... 10 runs averaged 0.0546228170395 seconds
... 10 runs averaged 0.0791359186172 seconds
```

### DIFF
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -393,7 +393,6 @@ cpdef write_map(bytearray fo, object datum, dict schema):
         write_long(fo, 0)
 
 
-
 INT_MIN_VALUE = -(1 << 31)
 INT_MAX_VALUE = (1 << 31) - 1
 LONG_MIN_VALUE = -(1 << 63)
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -387,7 +387,7 @@ cpdef write_map(bytearray fo, object datum, dict schema):
         # data. In this case, bail out so we don't possibly corrupt the output
         # file.
         if d_datum is not None:
-            raise
+            raise  # re-raise where d_datum is not None
 
         # Slower, general-purpose code where datum is something besides a dict,
         # e.g. a collections.OrderedDict or collections.defaultdict.
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -359,7 +359,7 @@ cpdef write_array(bytearray fo, list datum, schema):
     write_long(fo, 0)
 
 
-cpdef write_map(bytearray fo, dict datum, dict schema):
+cpdef write_map(bytearray fo, object datum, dict schema):
     """Maps are encoded as a series of blocks.
 
     Each block consists of a long count value, followed by that many key/value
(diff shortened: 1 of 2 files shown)


barrywhart:
I'm pretty sure the same technique could be extended to address #76 as well. Performance for the basic `dict` case would not be affected because the additional checks for `list` or `tuple` would happen if and only after the initial check for `dict` had failed.

fastavro/_write.pyx
```diff
@@ -521,21 +543,43 @@ cpdef write_union(bytearray fo, datum, schema):
...
+    cdef dict d_datum = None
     fields = schema['fields']
-    for field in fields:
-        name = field['name']
-        if name not in datum and 'default' not in field and\
-                'null' not in field['type']:
-            raise ValueError('no value and no default for %s' % name)
-        write_data(fo, datum.get(
-            name, field.get('default')), field['type'])
+    try:
```
scottbelden: If we wrote this `try` block like the following I don't think there would be a need for the test case that checks the stack trace (as I think it should be the same):

```python
    try:
        d_datum = <dict?>(datum)
    except TypeError:
        # Slower, general-purpose code where datum is something besides a dict,
        # e.g. a collections.OrderedDict or collections.defaultdict.
        for field in fields:
            name = field['name']
            if name not in datum and 'default' not in field and \
                    'null' not in field['type']:
                raise ValueError('no value and no default for %s' % name)
            write_data(fo, datum.get(
                name, field.get('default')), field['type'])
    else:
        # Faster, special-purpose code where datum is a Python dict.
        for field in fields:
            name = field['name']
            if name not in d_datum and 'default' not in field and \
                    'null' not in field['type']:
                raise ValueError('no value and no default for %s' % name)
            write_data(fo, d_datum.get(
                name, field.get('default')), field['type'])
```

At least it seemed the same when I did some quick testing...

fastavro/_write.pyx
```diff
@@ -369,13 +369,35 @@ cpdef write_map(bytearray fo, dict datum, dict schema):
...
     count in this case is the absolute value of the count written."""
-    if len(datum) > 0:
-        write_long(fo, len(datum))
-        vtype = schema['values']
-        for key, val in iteritems(datum):
-            write_utf8(fo, key)
-            write_data(fo, val, vtype)
-    write_long(fo, 0)
+    cdef dict d_datum = None
+    try:
```
scottbelden: Same comment about re-arranging the `try/except`.

fastavro/_write.pyx
barrywhart: Oh yeah, this looks like a good situation for `try/except/else` -- I didn't think of that. I'll try that out and make sure the generated code looks reasonable, performance wise. In theory, straight-line execution is fastest so it'd be nice not to have the `except` in the middle, but I think that's likely to be negligible impact here.

barrywhart: I went ahead and cleaned up the library code as you suggested. I'd like to keep the stack trace tests, even though they are kind of ugly and "white box", because they do help protect against future regression bugs where `dict` and other types end up using the same, slower code.

barrywhart: (BTW, the change you suggested did involve updating the stack trace tests, in case that was not clear from the above comment. They are still quite similar to before, only the specific stack trace checks changed a bit.)

Issue #144, opened 2018-01-18
Follow documented best practices for distributing Cython modules


barrywhart:
Issue #142 reminded me of a couple Cython packaging changes we should consider:
* Distribute `.c` files as well as `.pyx` files
* Do not enable Cython compilation by default

Both these practices are recommended in the [Cython documentation](http://cython.readthedocs.io/en/latest/src/reference/compilation.html#distributing-cython-modules).

The advantage is that it results in a less complex installation process on the target machine (just compiling C code instead of installing Cython and generating C code). It avoids any potential Cython version incompatibilities as well (i.e. if two different packages require different versions of Cython).

I'd be happy to help out making these changes if someone (@scottbelden?) can tell me how you currently prepare source packages for upload to PyPI.

scottbelden:
I believe the `.c` files should be in the source distribution. I just downloaded the 0.17.2 tarball from pypi and it has the c files. As far as how packing is done, it's via `make publish` which just calls the `publish.sh` script. I think this issue is related to https://github.com/tebeka/fastavro/pull/145

barrywhart:
One thing I wasn't sure about, when I ran `python setup.py sdist`, it would include the `.c` files if they already existed, but if they didn't, it wouldn't create them. Maybe `publish.sh` takes care of this. We'd also want to verify that if the `.c` files are present in the archive, it uses them instead of recreating them, etc. I probably should've checked all this before creating an issue, sorry about that!

scottbelden:
Resolved via https://github.com/tebeka/fastavro/pull/157

PR #145, opened 2018-01-19
Add wheel publishing to `publish.sh`

pbabics:
Hi, lately installing `fastavro` is taking quite a bit time, because it is compiling using `Cython`.

To faster up entire package installation you could build wheels (https://packaging.python.org/tutorials/distributing-packages/#wheels).

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -12,6 +12,8 @@ set -x
 
 make
 python setup.py sdist
+python setup.py bdist_wheel
+twine upload dist/fastavro-${ver}-*.whl
 twine upload dist/fastavro-${ver}.tar.gz
 # TODO: upload fails for some reason
 #conda build .
@@ -21,4 +23,5 @@ git push
 git push --tags
 # print sha so we can use it in conda-forge recipe
 sha256sum dist/fastavro-${ver}.tar.gz
+sha256sum dist/fastavro-${ver}-*.whl
 rm -fr build dist fastavro.egg-info/


scottbelden:
Thanks for the PR. Currently I run the publish script manually on OSX, so I'm guessing this would only be able to build/upload the OSX wheels, right? I think this might require some more tweaking (and some help from CI) if we wanted to have wheels built for all platforms, right?

I think https://github.com/tebeka/fastavro/issues/144 is somewhat related to this.

barrywhart:
Right, wheels include shared libraries. On Mac, you could use Docker to build Linux wheels. I use Docker at work and could possibly help.

scottbelden:
Yeah, that sounds like it could work for mac and linux. I've used docker plenty and so that shouldn't be too bad.

scottbelden:
@pbabics What platform do you run on?

pbabics:
Linux, I can prepare pipeline for Travis CI which would build wheel for linux and probably also mac

scottbelden:
@pbabics I made some changes to the `publish.sh` script so that it will check for and upload the wheels built for windows (which caused conflicts here). If you'd like to make changes to the travis steps so that it builds and saves the wheels for those platforms then we could have the publish script also upload those. Bonus points of the publish script can automatically download those wheels, but if that needs to be manual then that's fine (the windows wheels are currently downloaded manually as I couldn't find a good way to download them automatically from appveyor).

scottbelden:
Thanks for the info. I'll play around with this some and see if I can get it uploading the wheels via travis. It seems like it shouldn't be too difficult.

scottbelden:
I tried updating the travis file to build the wheels, but the upload failed. It turns out the solution isn't as easy as Windows and Mac and requires some of the work done for [manylinux](https://github.com/pypa/manylinux). I'll continue playing with it, but I'm not sure when it will be working fully.

pbabics:
Hi,

I was playing around a bit, and wrote a little "bash" one-liner, that builds wheels on `manylinux` image
```
docker run -v "$(pwd):/data" -it  quay.io/pypa/manylinux1_x86_64 bash -c 'cd /data; for PYBIN in /opt/python/*/bin; do ${PYBIN}/pip install cython; ${PYBIN}/python setup.py bdist_wheel; done'
```
This could be run in `before_deploy` script to build wheels and then using `pypi` deploy to publish them,
although I'm not sure if it will publish files which were produced before it, as it seems that `pypi` deploy calls `sdist` (and others builds if configured) itself.

Another option could be to build wheels and send them to `Github Releases` (https://docs.travis-ci.com/user/deployment/releases/) from which you could download them a upload to pypi using you `publish.sh` script manually.

scottbelden:
I made some modifications to the travis script to incorporate the docker building of the linux wheels. I tested this with a dummy tag (that I've since deleted) and it published the wheels to github which I'll plan to grab before uploading to pypi (like what is done with the windows wheels). The next time a release happens there should be linux wheels with it.

I'm going to close this now, but thanks for all the help.

kkirsanov:
Great work!
Thanks

Issue #148, opened 2018-01-23
fastavro(1) - Non-pretty failed due to indent=None


pyk:
I have an avro writer in `python3.6` like the following:

```python
from fastavro import writer

def main():
    output = "samples.snappy.avro"
    out = open(output, "wb")
    schema = {
        "doc": "test",
        "name": "Record",
        "namespace": "collection",
        "type": "record",
        "fields": [
            {"name": "key", "type": ["string", "null"]}
        ]
    }
    records = [
        {"key": None},
        {"key": "value"}
    ]
    writer(out, schema, records, codec="snappy")
    out.close()
    print("Generated: {}".format(output))

if __name__ == "__main__":
    main()

```

When I try to dump the generated avro file, it raise an error like the following:

```text
% fastavro samples.snappy.avro 
Traceback (most recent call last):
  File "/home/pyk/.virtualenvs/x/bin/fastavro", line 11, in <module>
    sys.exit(main())
  File "/home/pyk/.virtualenvs/x/lib/python3.6/site-packages/fastavro/__main__.py", line 77, in main
    json_dump(record, indent)
  File "/home/pyk/.virtualenvs/x/lib/python3.6/site-packages/fastavro/six.py", line 33, in py3_json_dump
    json.dump(obj, stdout, indent=indent)
TypeError: an integer is required (got type NoneType)
```

But when I use `--pretty`, it runs as expected:

```
% fastavro -p samples.snappy.avro 
{
    "key":null
}
{
    "key":"value"
}
```

I think the problem is on this line https://github.com/tebeka/fastavro/blob/master/fastavro/__main__.py#L73

pyk:
Oh yeah, I got this issue when listing a supported codecs

```
% fastavro --codecs
Traceback (most recent call last):
  File "/home/pyk/.virtualenvs/x/bin/fastavro", line 11, in <module>
    sys.exit(main())
  File "/home/pyk/.virtualenvs/x/lib/python3.6/site-packages/fastavro/__main__.py", line 50, in main
    print('\n'.join(sorted(avro._reader.BLOCK_READERS)))
AttributeError: module 'fastavro' has no attribute '_reader'

```

scottbelden:
Thanks for the reports. https://github.com/tebeka/fastavro/pull/149 should resolve both of these issues and add some testing that wasn't being done.

PR #149, opened 2018-01-23
test some of the fastavro __main__ functionality

scottbelden:
Resolves https://github.com/tebeka/fastavro/issues/148

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,6 @@ python:
     - pypy
     - pypy3
 install:
-    - pip install pytest flake8 check-manifest Cython
+    - pip install pytest flake8 check-manifest Cython ujson
 script:
     - ./run-tests.sh
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ environment:
         - PYTHON: "C:\\Python36-x64"
 
 install:
-    - "%PYTHON%\\python.exe -m pip install pytest flake8 check-manifest Cython wheel"
+    - "%PYTHON%\\python.exe -m pip install pytest flake8 check-manifest Cython wheel ujson"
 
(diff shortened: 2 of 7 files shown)


scottbelden:
@barrywhart I was hoping to get your opinion on removing the `_six.pyx` module. At the moment the two `six` modules are basically the same (except where `six.py` was updated and `_six.pyx` wasn't (https://github.com/tebeka/fastavro/blob/master/fastavro/six.py#L71-L73)).

I'm not sure that there's a reason to have both if they are both the same but wanted to see if you had some insight that I'm missing.

barrywhart:
@scottbelden: You may recall, before we had separate files for Cython, the only Cython-specific was this line at the top of some files (including `six.py`):

```
# cython: auto_cpdef=True
```

While other Cython files were updated to take more advantage of Cython, `_six.pyx` was not.

You're probably right that Cython isn't helping *much* here, but it may be having some impact because of `auto_cpdef`. In some cases, I have found that simply compiling a Python file to Cython and/or declaring functions as `cpdef` (with no other changes) makes a small speed improvement because the code is being converted statically to C, eliminating some CPython interpreter overhead (e.g. bytecode interpretation). In some cases, I've seen speedups of 30% (this was with fairly complex code, similar to fastavro's `_read.pyx` and `_write.pyx` code). Given how simple `_six.py` is (mostly one-line functions that forward to other functions, or even just aliases), the speedup is likely to be much less than this or even negligible.

I suggest eliminating the file and running some read and write benchmarks. If you don't notice a meaningful difference, then go ahead with eliminating it.

scottbelden:
From some quick runs of the benchmark, I think having the `_six.pyx` file does help, so I've kept it.

PR #153, opened 2018-01-25
Write union using tuple with schema name

Artimi:
Hi,
I've noticed cool feature in `write_union` that allow to specify schema of records precisely instead of guessing based on parameters. However this does not allow to use acquainted schemata (because string is not record and therefore has no `name` to look for). I think that this change will increase possibility of the feature. Please take a look. I will be happy to change the code according to your suggestions.
P.S. : I didn't create issue, should I?

### DIFF
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -502,12 +502,16 @@ cpdef write_union(bytearray fo, datum, schema):
     cdef int32 most_fields
     cdef int32 index
     cdef int32 fields
+    cdef str schema_name = None
     if isinstance(datum, tuple):
         (name, datum) = datum
         for index, candidate in enumerate(schema):
             if extract_record_type(candidate) == 'record':
-                if name == candidate["name"]:
-                    break
+                schema_name = candidate["name"]
+            else:
+                schema_name = candidate
+            if name == schema_name:
+                break
         else:
             msg = 'provided union type name %s not found in schema %s' \
                 % (name, schema)
--- a/fastavro/_write_py.py
+++ b/fastavro/_write_py.py
@@ -385,8 +385,11 @@ def write_union(fo, datum, schema):
         (name, datum) = datum
         for index, candidate in enumerate(schema):
             if extract_record_type(candidate) == 'record':
-                if name == candidate["name"]:
-                    break
+                schema_name = candidate['name']
+            else:
+                schema_name = candidate
+            if name == schema_name:
+                break
         else:
             msg = 'provided union type name %s not found in schema %s' \
                 % (name, schema)
--- a/tests/test_fastavro.py
+++ b/tests/test_fastavro.py
@@ -1237,3 +1237,42 @@ def test_regular_vs_ordered_dict_map_typeerror():
     # be different, indicating the exception occurred at a different line
     # number.
     assert filtered_stacks[0] != filtered_stacks[1]
+
+
+def test_write_union_tuple_primitive():
+    '''
+    Test that when we can use tuple style of writing unions
+    (see function `write_union` in `_write`) with primitives
+     not only with records.
+    '''
+
+    schema = {
+        'name': 'test_name',
+        'namespace': 'test',
+        'type': 'record',
+        'fields': [
+            {
+                'name': 'val',
+                'type': ['string', 'int']
+            }
+        ]
+    }
+
+    data = [
+        {"val": ("int", 1)},
+        {"val": ("string", "string")},
+    ]
+
+    expected_data = [
+        {"val": 1},
+        {"val": "string"},
+    ]
+
+    new_file = MemoryIO()
+    fastavro.writer(new_file, schema, data)
+    new_file.seek(0)
+
+    new_reader = fastavro.reader(new_file)
+    new_records = list(new_reader)
+
+    assert new_records == expected_data


scottbelden:
Thanks for the PR. No need to create an issue. If I understand correctly, it sounds like you have a case where there is a field that is a union of types such as `int`, `string`, etc. and you want to be able to have your tuple specify that the type is `string` but that's not possible because currently it expects the candidate schema to be a `record` type.

If so, the code changes look correct, but the test seems like it's not testing the new functionality. I would expect the test to be more like the following:

```python
def test_write_union_tuple_acquainted_schemata():
    '''
    Test that when we are writing unions using acquainted schemata we can use
    tuple style of writing unions (see function `write_union` in `_write`).
    '''

    schema = {
        'name': 'test_name',
        'namespace': 'test',
        'type': 'record',
        'fields': [
            {
                'name': 'val',
                'type': ['string', 'int']
            }
        ]
    }

    data = [
        {"val": ("int", 1)},
        {"val": ("string", "string")},
    ]

    expected_data = [
        {"val": 1},
        {"val": "string"},
    ]

    new_file = MemoryIO()
    fastavro.writer(new_file, schema, data)
    new_file.seek(0)

    new_reader = fastavro.reader(new_file)
    new_records = list(new_reader)

    assert new_records == expected_data
```

Is that right? Or am I misunderstanding the reason?

Artimi:
@scottbelden Yes, that is exactly what I want to achieve. Maybe I took overly complicated example but it reflects our needs exactly. I can depict it a bit.

We are encoding some general message that is only union of other data types plus some metadata. Those datatypes are records that can even have the same fields but it is important to distinguish what type it really is because it carries infomation too. Basically we have something like this (just dummy example):
```
{
	"name": "Create",
	"type": "record",
	"fields": [{
		"name": "data",
		"type": "string"
	}]
}

{
	"name": "Cancel",
	"type": "record",
	"fields": [{
		"name": "data",
		"type": "string"
	}]
}

{
	"name": "Message",
	"type": "record",
	"fields": [{
		"name": "metadata",
		"type": "string"
	}, {
		"name": "data",
		"type": ["Create", "Cancel"]
	}]
}
```
So there is no way for fastavro to tell which schema will be used and it would encode everything as `Create` probably. So I need to set correct schema on writer side.
On reader side there was a problem how to get the schema but I patched function `read_union` so it set attribute to the resulting dict with specified schema. 
I probably could encode this information as another enum that would say the type. But it seem to me as wasting the space and sending redundant information because we already may know what schema was used.

So the test is exactly suited for our needs. But I understand it may seem overly complicated. I've changed it to your example.

scottbelden:
Sorry. I'm confused. What you explained about having two records that are the same except in their name and wanting to make sure they are encoded correctly should already be possible.

Did your original test fail without the code changes? I should have checked that, but I didn't and now the commit has been changed so I'm having trouble finding that original commit.

Artimi:
Yes original test failed without code changes. Original test looked like this:
```python
def test_write_union_tuple_acquainted_schemata():
    '''
    Test that when we are writing unions using acquainted schemata we can use
    tuple style of writing unions (see function `write_union` in `_write`).
    '''
    fastavro.schema.acquaint_schema({
        "type": "record",
        "name": "A",
        "namespace": "test",
        "fields": [{
            "name": "field",
            "type": "int",
        }]
    })
 
    fastavro.schema.acquaint_schema({
        "type": "record",
        "name": "B",
        "namespace": "test",
        "fields": [{
            "name": "field",
            "type": "string",
        }]
    })
 
    schema = {
        'name': 'test_name',
        'namespace': 'test',
        'type': 'record',
        'fields': [
            {
                'name': 'val',
                'type': ["test.A", "test.B"]
            }
        ]
    }
 
    data = [
        {"val": ("test.A", {"field": 1})},
        {"val": ("test.B", {"field": "string"})},
    ]
 
    expected_data = [
        {"val": {"field": 1}},
        {"val": {"field": "string"}},
    ]
 
    new_file = MemoryIO()
    fastavro.writer(new_file, schema, data)
    new_file.seek(0)
 
    new_reader = fastavro.reader(new_file)
    new_records = list(new_reader)
 
    assert new_records == expected_data
```
And it failed with 
```
  File "/home/sebekpet/projects/fastavro/tests/test_fastavro.py", line 1290, in test_write_union_tuple_acquainted_schemata
    fastavro.writer(new_file, schema, data)
  File "fastavro/_write.pyx", line 854, in fastavro._write.writer
    output.write(record)
  File "fastavro/_write.pyx", line 777, in fastavro._write.Writer.write
    write_data(self.io.value, record, self.schema)
  File "fastavro/_write.pyx", line 635, in fastavro._write.write_data
    return fn(fo, datum, schema)
  File "fastavro/_write.pyx", line 539, in fastavro._write.write_record
    cpdef write_record(bytearray fo, object datum, dict schema):
  File "fastavro/_write.pyx", line 567, in fastavro._write.write_record
    write_data(fo, d_datum.get(
  File "fastavro/_write.pyx", line 635, in fastavro._write.write_data
    return fn(fo, datum, schema)
  File "fastavro/_write.pyx", line 496, in fastavro._write.write_union
    cpdef write_union(bytearray fo, datum, schema):
  File "fastavro/_write.pyx", line 514, in fastavro._write.write_union
    raise ValueError(msg)
ValueError: provided union type name test.A not found in schema ['test.A', 'test.B']
```
The problem is that `write_union` check if each `candidate` is record which is not true when it is a type (i.e. acquainted schema). 

```python
    if isinstance(datum, tuple):
        (name, datum) = datum
        for index, candidate in enumerate(schema):
            if extract_record_type(candidate) == 'record':
                if name == candidate["name"]:
                    break
        else:
            msg = 'provided union type name %s not found in schema %s' \
                % (name, schema)
            raise ValueError(msg)
```

What I've added to the code is that checks not only records name but if it is not record it also test schema as is.

scottbelden:
Thanks. Now I have a grasp of what's going on.

I'm tempted to merge what you have currently since it does fix an issue of using the tuple selector for primitive types. It also as a side effect works in your case if you use the `acquaint_schema` like you did. However, it's not very robust. For example, I think your original test should more accurately be written as the following:

```python
def test_write_union_tuple_acquainted_schemata():
    '''
    Test that when we are writing unions using acquainted schemata we can use
    tuple style of writing unions (see function `write_union` in `_write`).
    '''

    schema = {
        'name': 'test_name',
        'namespace': 'test',
        'type': 'record',
        'fields': [
            {
                'name': 'val',
                'type': [
                    {
                        'name': 'A',
                        'namespace': 'test',
                        'type': 'record',
                        'fields': [
                            {'name': 'field', 'type': 'int'},
                        ],
                    },
                    {
                        'name': 'B',
                        'namespace': 'test',
                        'type': 'record',
                        'fields': [
                            {'name': 'field', 'type': 'string'},
                        ],
                    }
                ]
            }
        ]
    }

    data = [
        {"val": ("test.A", {"field": 1})},
        {"val": ("test.B", {"field": "string"})},
    ]

    expected_data = [
        {"val": {"field": 1}},
        {"val": {"field": "string"}},
    ]

    new_file = MemoryIO()
    fastavro.writer(new_file, schema, data)
    new_file.seek(0)

    new_reader = fastavro.reader(new_file)
    new_records = list(new_reader)

    assert new_records == expected_data
```

However, even with these changes the test would fail with the following:

```python
  File "/Users/scott/git/fastavro/tests/test_fastavro.py", line 1288, in test_write_union_tuple_acquainted_schemata
    fastavro.writer(new_file, schema, data)
  File "/Users/scott/git/fastavro/fastavro/_write_py.py", line 675, in writer
    output.write(record)
  File "/Users/scott/git/fastavro/fastavro/_write_py.py", line 598, in write
    write_data(self.io, record, self.schema)
  File "/Users/scott/git/fastavro/fastavro/_write_py.py", line 491, in write_data
    return fn(fo, datum, schema)
  File "/Users/scott/git/fastavro/fastavro/_write_py.py", line 434, in write_record
    name, field.get('default')), field['type'])
  File "/Users/scott/git/fastavro/fastavro/_write_py.py", line 491, in write_data
    return fn(fo, datum, schema)
  File "/Users/scott/git/fastavro/fastavro/_write_py.py", line 387, in write_union
    (name, datum) = datum
ValueError: provided union type name test.A not found in schema [{'name': 'A', 'namespace': 'test', 'fields': [{'name': 'field', 'type': 'int'}], 'type': 'record'}, {'name': 'B', 'namespace': 'test', 'fields': [{'name': 'field', 'type': 'string'}], 'type': 'record'}]
```

I think maybe a better solution would be for `write_union` to do something like the following:

```python
def write_union(fo, datum, schema):
    """A union is encoded by first writing a long value indicating the
    zero-based position within the union of the schema of its value. The value
    is then encoded per the indicated schema within the union."""

    if isinstance(datum, tuple):
        (name, datum) = datum
        for index, candidate in enumerate(schema):
            if extract_record_type(candidate) == 'record':
                schema_name = get_namespaced_name(candidate)  # <--- new function
            else:
                schema_name = candidate
            if name == schema_name:
                break
        else:
            msg = 'provided union type name %s not found in schema %s' \
                % (name, schema)
            raise ValueError(msg)
```

Then it should work regardless of whether the `candidate` schema is `test.A` (as it is in your example) or a full schema definition (as it is in my example). However, getting the full namespaced name is a little tricky... We do it currently as a part of processing the schema, but I don't think we have a good mapping of the namespaced name to the schema definition.

All in all, if you okay with this PR as is and it helps you, I'm good with merging it. I will plan to make a ticket to tackle the problem of making the code a little more robust in the future.

Artimi:
@scottbelden You are right, it is not very robust. Actually there is such function that you named`get_namespaced_name`: in `schema` is function `schema_name` that adds `namespace` from `record` if present. Your test is passing when I made few changes in the code. Take a look. I'm not sure if I should add `schema_name` to `schema.py` and therefore expose it's api. It works without it so maybe there is no need, but btw imports in fastavro are quite magic.

scottbelden:
I saw the `schema_name` function, but it doesn't help much. It would help in the case where the namespace is specified, but if the namespace is in the parent schema, it won't work. Given that, I think it would be best to remove the most recent commit and just merge the first commit (if that will still solve your immediate issue).

Artimi:
Ok done.

scottbelden:
Thanks!

PR #157, opened 2018-02-07
When installing from source package, use the included .c files

barrywhart:
PR for #144: Follow the [recommendations for distributing Cython packages](http://cython.readthedocs.io/en/latest/src/reference/compilation.html#distributing-cython-modules).

`setup.py` now checks for an environment variable `FASTAVRO_USE_CYTHON` when it runs. (Note: If it seems better, we could instead use a command-line option or some other technique.) Some techniques for using non-standard command-line options with `setup.py` are covered [here](https://stackoverflow.com/questions/677577/distutils-how-to-pass-a-user-defined-parameter-to-setup-py).

If the environment variable is set to `1` (which should be the case when running `setup.py develop` and during packaging/publishing), the library requires Cython and uses the `.pyx` files. If not set (which should be the case during installation), it does not require Cython and simply compiles the `.c` files which are assumed to be included with the package.

Benefits to users include:
* Faster installation
  * Does not download or install Cython
  * Does not have to convert `.pyx` files to `.c` files
* Avoids any potential Cython version issues with other packages used in the user's application. (We've seen versions of Cython that don't work with `fastavro` and certain versions of CPython, etc.)

To test these changes:
```
make
FASTAVRO_USE_CYTHON=1 python setup.py sdist
pushd ..
pyenv deactivate
pyenv uninstall -f fastavro-test
pyenv virtualenv fastavro-test
pyenv activate fastavro-test
pip install fastavro/dist/fastavro-0.17.7.tar.gz  # Or whatever the current version is
```

### DIFF
--- a/.travis_build_wheel.sh
+++ b/.travis_build_wheel.sh
@@ -13,4 +13,4 @@ TRAVIS_PYTHON_VERSION=$1
 TRAVIS_PYTHON_VERSION=${TRAVIS_PYTHON_VERSION/.}
 
 /opt/python/*${TRAVIS_PYTHON_VERSION}*m/bin/pip install cython
-/opt/python/*${TRAVIS_PYTHON_VERSION}*m/bin/python setup.py bdist_wheel
+FASTAVRO_USE_CYTHON=1 /opt/python/*${TRAVIS_PYTHON_VERSION}*m/bin/python setup.py bdist_wheel
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,11 +18,13 @@ build: off
 
 test_script:
     - "set PYTHONPATH=%cd%"
+    - "set FASTAVRO_USE_CYTHON=1"
     - "%PYTHON%\\python.exe setup.py build_ext --inplace"
     - "%PYTHON%\\python.exe -m pytest -v tests"
 
 after_test:
     # If tests are successful, create binary packages for the project.
+    - "set FASTAVRO_USE_CYTHON=1"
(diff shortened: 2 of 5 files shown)


scottbelden:
Awesome. This looks good to me. I did add some new stuff to the travis config so that it calls the `.travis_build_wheels.sh` file to build linux wheels. I'm assuming we need to add the `FASTAVRO_USE_CYTHON=1` in [here](https://github.com/tebeka/fastavro/blob/master/.travis_build_wheel.sh#L16) as well?

barrywhart:
That's right. You'll need `FASTAVRO_USE_CYTHON=1` anywhere the `setup.py` file needs to translate `.pyx` files to `.c` files.

An alternate way of doing this would be to apply the `publish.sh` technique everywhere: That is, run the `cython` command-line tool to do the `.c` translation before running `setup.py`. In that case, we could remove any mention of Cython from `setup.py`. That actually seems simpler and cleaner to me. I did it the other way (in the `setup.py` file) since that's what the Cython people suggest, but TBH I don't see much value to `fastavro` users from doing that.

scottbelden:
I think the way it's done here is good. I'd be a little hesitant to do things in the publish script because 1) we would need a separate one for windows and 2) I think having the publish be a manual process isn't a bad thing.

barrywhart:
Sounds good to me. `setup.py` scripts can be funky, but a relatively standardized kind of funky. ¯\_(ツ)_/¯

PR #158, opened 2018-02-10
Update README and package classifiers

barrywhart:
This PR has some proposed changes to the `README.md` and the classifiers in the `setup.py`. I feel like the `fastavro` package is selling itself short a bit in terms of maturity, e.g. listing itself as "beta" and asking for people to help maintain it. By updating this a bit, people will be more likely to use it on their projects.

In addition to some other changes, I removed some of the info about Cython -- I think that's a technical detail we should omit as it may scare people off, especially now that Cython is not required to be installed on the target machine (only a C compiler). Other libraries that use Cython (e.g. `scipy`) do not draw attention to its use.


scottbelden:
Thanks!

Issue #169, opened 2018-03-30
Missing linux wheels


pbabics:
Hi,

I was looking for wheels built for linux and it seems that they are not uploaded to pypi.
Could you please upload them from releases (https://github.com/tebeka/fastavro/releases) ?

PR #170, opened 2018-03-30
Add checks for linux wheels to `publish.sh`

pbabics:
This should prevent missing linux wheels, see https://github.com/tebeka/fastavro/issues/169


scottbelden:
Thanks

PR #174, opened 2018-04-25
Fastavro 171 logical timestamp tz support

lsterk:
Fix for Issue 171 (adding in tzaware datetime objects)

One thing to note is that the output Python datetime object is now TZ-aware. TZ-aware datetime objects can't be compared with TZ-naive objects, so some downstream users might have problems if they rely on datetime objects being TZ-naive. However, the code branch for a Python2.7 (i.e. no datetime.timestamp() method) remains unchanged for TZ-naive objects when serializing the object as a long for Avro encoding.

Also, sorry if I missed any prep for the PR, it seemed like the "Hacking" section was talking about after a PR to fastavro gets merged... so I didn't do any of those steps. Lemme know if there's stuff I need to do from that section.

### DIFF
--- a/fastavro/_timezone.py
+++ b/fastavro/_timezone.py
@@ -1,30 +1,16 @@
-"""Implementation of abstract base class tzinfo"""
-# Licensed to the Apache Software Foundation (ASF) under one
-# or more contributor license agreements.  See the NOTICE file
-# distributed with this work for additional information
-# regarding copyright ownership.  The ASF licenses this file
-# to you under the Apache License, Version 2.0 (the
-# "License"); you may not use this file except in compliance
-# with the License.  You may obtain a copy of the License at
-#
-# http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
+"""Timezone class for dealing with timezone-aware datetime objects
+Inspired by https://github.com/apache/avro/pull/207"""
 import time
 from datetime import datetime
 from datetime import timedelta
 from datetime import tzinfo
 
-"""Inspired by https://github.com/apache/avro/pull/207
-Obsoleted by datetime.timezone in py3.2"""
+
 
 
 class UTCTzinfo(tzinfo):
-
+    """Implementation of abstract base class tzinfo,
+    python >= 3.2 can use datetime.timezone.utc instead"""
     def utcoffset(self, dt):
         return timedelta(0)
 
@@ -38,12 +24,5 @@ class UTCTzinfo(tzinfo):
 utc = UTCTzinfo()
 
 # used to compute timestamp for tz-aware datetime objects
+# python >= 3.3 can use datetime.datetime.timestamp() instead
 epoch = datetime(1970, 1, 1, tzinfo=utc)
-
-
-def assert_naive_datetime_equal_to_tz_datetime(naive_datetime, tz_datetime):
-    # mktime appears to ignore microseconds, so do this manually
-    timestamp = int(time.mktime(naive_datetime.timetuple()))
-    timestamp += float(naive_datetime.microsecond) / 1000 / 1000
-    aware_datetime = datetime.fromtimestamp(timestamp, tz=utc)
-    assert aware_datetime == tz_datetime
--- a/fastavro/_timezone.py
+++ b/fastavro/_timezone.py
@@ -1,0 +1,49 @@
+"""Implementation of abstract base class tzinfo"""
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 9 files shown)


fastavro/_timezone.py
```diff
@@ -0,0 +1,49 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import time
+from datetime import datetime
+from datetime import timedelta
+from datetime import tzinfo
+
+"""Inspired by https://github.com/apache/avro/pull/207
+Obsoleted by datetime.timezone in py3.2"""
```
scottbelden: To make sure I'm clear, if/when python 2 support is dropped, we should be able to remove `UTCTzinfo` and just use `datetime.timezone.utc`, correct?

fastavro/_timezone.py
```diff
@@ -0,0 +1,49 @@
...
+        return timedelta(0)
+
+
+utc = UTCTzinfo()
+
+# used to compute timestamp for tz-aware datetime objects
+epoch = datetime(1970, 1, 1, tzinfo=utc)
+
+
+def assert_naive_datetime_equal_to_tz_datetime(naive_datetime, tz_datetime):
```
scottbelden: Can we move this function to the test directory? It can live under https://github.com/tebeka/fastavro/blob/master/tests/conftest.py as it seems to be shared between multiple test files.

tests/test_timezone.py
```diff
@@ -0,0 +1,208 @@
...
+              "type": {'type': 'long', 'logicalType': 'timestamp-micros'}
+          },
+    ],
+    "namespace": "namespace",
+    "name": "name",
+    "type": "record"
+}
+
+
+# Test Time Zone with fixed offset and 1 hr DST
```
scottbelden: I think either this comment or the one for the `TDTTzinfo` is incorrect? They are both the same but the classes are different.

abrarsheikh:
lg2m, ship it

fastavro/_timezone.py
lsterk: Correct. That's exactly right.

fastavro/_timezone.py
```diff
@@ -0,0 +1,49 @@
+"""Implementation of abstract base class tzinfo"""
+# Licensed to the Apache Software Foundation (ASF) under one
```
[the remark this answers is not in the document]
lsterk: ++, the boilerplate escaped my notice. My first pass at this had a lot more in _timezone (from a PR in apache/avro) but this revision is much simpler.

fastavro/_timezone.py
lsterk: That's definitely reasonable. AFAIK there isn't any magical sharing of non-fixtures from ```conftest.py``` -- I will still need to do ```from conftest import assert_naive_datetime...``` in every test using this.

I'll have a new revision with this change.

tests/test_timezone.py
lsterk: ++. 

Side note: from [Python 3 API](https://docs.python.org/3/library/datetime.html#datetime.timezone.dst) it looks like DST was something that kinda got dropped in Python 3 (i.e. always return None). By contrast, [Python 2 API](https://docs.python.org/2/library/datetime.html#datetime.tzinfo.dst) it seems that when timezones were just an Abstract Base Class, there was some notion of useful information from DST. 

Given this, I think it's worth removing any test for daylight savings time compatibility given that this seems to be kinda unsupported in Python3. I'm open to feedback here though.

kkirsanov:
I checked on my test data. That is all right.

I just wanted to clarify, why make a new class for timezone (UTCTzinfo) when there is a ready pytz.UTC?

fastavro/_timezone.py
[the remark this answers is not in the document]
scottbelden: Do you mind removing it in that case?

fastavro/_timezone.py
scottbelden: Yeah, there is no magic with non-fixtures, but the `from conftest` seems fine.

tests/test_timezone.py
scottbelden: That sounds fine to me. I'll defer to @kkirsanov and your judgements as I don't really use the timezone stuff.

lsterk:
@kkirsanov The main reason for making a new class for timezone is that pytz is kinda overbuilt for this. The only logic that this package needs is the definition of UTC; I think making a new class makes this very clear.

I read a few sources that suggested that pytz's extra features make it slightly less performant than other implementations. I opted to explicitly define UTC in a minimalistic way to reduce performance impact of these change.

That being said, I'm willing to give it a go using ```pytz``` but it seems like most benchmarks have it performing significantly slower (to be fair, mostly when converting between timezones, i.e. when writing a tz-aware datetime object into Avro).

lsterk:
@scottbelden thanks for your feedback! I look forward to seeing this released soon! 

@kkirsanov, possibly an even more important reason to avoid `pytz` is to that it avoids adding another dependency for such a simple feature. That was something that came up when we tried to merge a similar fix for `apache/avro`.

My benchmark remark was mostly based on [blog post](https://pendulum.eustace.io/blog/a-faster-alternative-to-pyz.html) for a competitor to pytz - so I guess it's worth taking that with a grain of salt. But keeping this change as simple as possible has its merits -- and explicitly defining the new feature instead of relying on an outside package makes this easier to review and merge.

Issue #176, opened 2018-04-28
decoding with compatible schema crashes with exception


abrarsheikh:
There appears to be a bug in the decoding logic when the reader schema is different from writer schema

**writer_schema**
```
    {
        'name': 'name1',
        'type': 'record',
        'namespace': 'namespace1',
        'fields': [{
            'doc': 'test',
            'type': 'int',
            'default': 1,
            'name': 'good_field'
        }], 
        'doc': 'test'
    }
```

**reader_schema**
```
    {
        'name': 'name1',
        'doc': 'test',
        'namespace': 'namespace1',
        'fields': [{
            'name': 'good_field',
            'doc': 'test',
            'default': 1,
            'type': 'int'
        }, {
            'name': 'good_compatible_field',
            'doc': 'test',
            'default': 1,
            'type': 'int'
        }],
        'type': 'record'
    }
```

when i try to use the `reader_schema` to decode the encoded_data with `writer_schema` i get following exception
```
Traceback (most recent call last):
  File "fast_avro_bug.py", line 40, in <module>
    'type': 'record'
  File "fastavro/_read.pyx", line 690, in fastavro._read.schemaless_reader
  File "fastavro/_read.pyx", line 692, in fastavro._read.schemaless_reader
  File "fastavro/_read.pyx", line 524, in fastavro._read.read_data
  File "fastavro/_read.pyx", line 536, in fastavro._read._read_data
  File "fastavro/_read.pyx", line 432, in fastavro._read.read_record
  File "fastavro/_read.pyx", line 454, in fastavro._read.read_record
  File "fastavro/_read.pyx", line 536, in fastavro._read._read_data
  File "fastavro/_read.pyx", line 216, in fastavro._read.read_long
  File "fastavro/_read.pyx", line 228, in fastavro._read.read_long
StopIteration
```

as per the avro spec those 2 schemas should be compatible as the field `good_compatible_field` has a `default` value of `1`.

**sample code**
```
import fastavro
from six import BytesIO

stringio = BytesIO()
fastavro.schemaless_writer(
    stringio,
    {
        'name': 'name1',
        'type': 'record',
        'namespace': 'namespace1',
        'fields': [{
            'doc': 'test',
            'type': 'int',
            'default': 1,
            'name': 'good_field'
        }], 
        'doc': 'test'
    },
    {'good_field': 1}
)
encoded_data = stringio.getvalue()

actual_data = fastavro.schemaless_reader(
    BytesIO(encoded_data),
    {
[...]

PR #180, opened 2018-05-02
allow for schema migration in the schemaless reader

scottbelden:
Resolves https://github.com/tebeka/fastavro/issues/176

### DIFF
--- a/fastavro/_read.pyx
+++ b/fastavro/_read.pyx
@@ -687,9 +687,11 @@ class reader:
 iter_avro = reader
 
 
-cpdef schemaless_reader(fo, schema):
-    acquaint_schema(schema)
-    return read_data(fo, schema)
+cpdef schemaless_reader(fo, writer_schema, reader_schema=None):
+    acquaint_schema(writer_schema)
+    if reader_schema:
+        populate_schema_defs(reader_schema)
+    return read_data(fo, writer_schema, reader_schema)
 
 
 cpdef is_avro(path_or_buffer):
--- a/fastavro/_read_py.py
+++ b/fastavro/_read_py.py
@@ -602,15 +602,18 @@ class reader:
 iter_avro = reader
 
 
-def schemaless_reader(fo, schema):
+def schemaless_reader(fo, writer_schema, reader_schema=None):
     """Reads a single record writen using the schemaless_writer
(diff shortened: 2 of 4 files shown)


scottbelden:
@abrarsheikh Let me know if this will work for you regarding https://github.com/tebeka/fastavro/issues/176. You will still need the original schema that was used to write out the avro binary, but now you can supply the new reader schema to handle new fields and other schema migration.

abrarsheikh:
@scottbelden thanks for picking this up, does this fully support [schema resolution](https://avro.apache.org/docs/1.8.1/spec.html#Schema+Resolution) with schemaless reader? 

can you point me to some docs that talk about the difference between  `writer`/`reader` and `schemaless_writer`/`schemaless_reader` ?

scottbelden:
The current released version of fastavro doesn't support schema resolution in the schemaless reader/writer. Schema resolution is supported with the normal reader/writer.

Typically when writing avro binary with the normal writer, it includes the standard avro header, metadata, and then finally all of the record data (for one or more records). To read the encoded data you don't need to provide the schema that was used to write the data because that schema was included in the binary output. Of course, if the schema has changed you would need to provide the new `reader_schema` so that proper resolution can be done.

In the case of the schemaless reader/writer, only the data is encoded, not all the other information like the schema, headers, etc. This means that when you go to read the data, you need to provide the schema that was used when it was encoded. Both sides must agree on what schema was used for the data. One of the use cases for something like this is two network endpoints that want to reduce the size of the data being transmitted between them (to conserve bandwidth). Since the headers and schema (which could be quite large) is not included, the total binary is smaller, but once again, both parties need to know what schema was used during writing.

This change would allow the `schemaless_reader` to supply an updated `reader_schema` (in addition to still needing to supply the original schema that was used for writing) to allow for schema migration/resolution.

abrarsheikh:
@scottbelden awesome. Thanks a lot for the context, i now have a better understanding of how this work. I am going to take a stab at this code review today

abrarsheikh:
lg2n, ship it

abrarsheikh:
thanks, is this feature released in `0.18.2`?

scottbelden:
Yes, it should be.

PR #182, opened 2018-05-08
Validators

rhaarm:
After the fact, I noticed that some of the int/float validators have datetime related stuff, those could potentially be pulled out into validating logicalTypes, but not sure right now. Just KIS at the moment.


scottbelden:
It might take me a few days to find some time to give this a proper look over but I'll try to do that soon. In the meantime, do you mind resolving the merge conflict? Thanks!

scottbelden:
Looks like it was just a post-test step of doing some docker stuff that we do for building wheels. It can be ignored.

fastavro/_write_common.py
```diff
@@ -0,0 +1,18 @@
...
+            format(datum=datum, given_type=type(datum),
+                   schema=schema, field=field)
+        super(ValidationException, self).__init__(message)
+        self.message = message
+        self.datum = datum
+        self.schema = schema
+        self.field = field
+
+
+class ValidationErrors(ValidationException):
```
scottbelden: Can we get away with not raising this exception to the user and only having the single `ValidationException` (or alternatively only having this exception not the other). I understand this is a subclass of that, but I can't imagine when a user would only want to catch this.

Or maybe `ValidationErrors` should be kept and the `__init__` could take in `*errors` where an `error` is some `namedtuple` or simple `ValidationError` class that has the `datum`, `schema`, and `field` property with a `__str__` that builds the message and the `ValidationErrors` exception can build the overall message based on the concatenation of the errors. Something like the following:

```python
class ValidationErrors(Exception):
    def __init__(self, *errors):
        super(Exception, self).__init__(', '.join(error for error in errors))
        self.errors = errors

class ValidationError(object):
    def __init__(self, datum, schema, field):
        self.datum = datum
        self.schema = schema
        self.field = field

    def __str__(self):
        return '{field} is {datum} of type {given_type} expected {schema}'. \
            format(datum=self.datum,
                   given_type=type(self.datum),
                   schema=self.schema,
                   field=self.field)
```

Alternatively, `ValidationErrors` could have some classmethod helpers like so:

```python
class ValidationErrors(Exception):
    
    @classmethod
    def from_single(cls, datum, schema, field):
        return ValidationErrors('{field} is {datum} of type {given_type} expected {schema}'. \
            format(datum=datum,
                   given_type=type(datum),
                   schema=schema,
                   field=field)
        )

    @classmethod
    def from_many(cls, messages):
        return ValidationErrors(', '.join(message for message in messages))
```

What are your thoughts?

fastavro/_write_py.py
```diff
@@ -296,89 +312,174 @@ def write_map(fo, datum, schema):
...
+        result = validate(datum,
+                          schema=SCHEMA_DEFS[record_type],
+                          field=ns_field,
+                          raise_errors=raise_errors)
+
+    if raise_errors and result is False:
+        raise ValidationException(datum, schema, ns_field)
+
+    if result is None:
+        raise UnknownType(record_type)
```
scottbelden: This is technically a breaking change as `validate` used to throw a `ValueError` when there was an unknown schema, but I like this better and I don't think there would be much to any impact. :+1:

fastavro/_write_py.py
```diff
@@ -296,89 +312,174 @@ def write_map(fo, datum, schema):
...
+            ret = validate(datum, schema=s,
+                           field=None,
+                           raise_errors=raise_errors)
+            if ret:
+                # We exit on the first passing type in Unions
+                return True
+        except ValidationException as e:
+            errors.append(e.message)
+    if raise_errors:
+        raise ValidationErrors(errors)
```
scottbelden: If we change the way the exceptions are used as mentioned in the big comment, this could be changed to:

```python
raise ValidationErrors(*errors)
```
or
```python
raise ValidationErrors.from_many(errors)
```

fastavro/_write_py.py
```diff
@@ -296,89 +312,174 @@ def write_map(fo, datum, schema):
...
+
+
+def validate(datum, schema, field=None, raise_errors=False):
+    """Determine if a python datum is an instance of a schema."""
+    record_type = extract_record_type(schema)
+    result = None
+
+    if hasattr(schema, 'get'):
+        if field is not None:
+            ns_field = '.'.join([field, schema.get('name')])
```
scottbelden: Either I'm not understanding the intention of this correctly, or I don't think it's working correctly. I changed your `test_validate_int_in_string_raises` to actually raise the exception like so:

```python
    # with pytest.raises((ValidationException,)):
    #     validation_raise(schema, *records)
    validation_raise(schema, *records)
```

And the result is `fastavro._write_common.ValidationException: missingerror is 11 of type <class 'int'> expected string`

If I understand correctly, I think the intention is that the message should say `missingerror.str`, not just `missingerror`

fastavro/_write_py.py
```diff
@@ -296,89 +312,174 @@ def write_map(fo, datum, schema):
...
+        else:
+            ns_field = schema.get('name')
+    else:
+        ns_field = field
+
+    if record_type in ('null', 'union'):
```
scottbelden: Can you explain why this block needs to first check `null` and `union` and then the `elif` and then the `else`?

If I change this entire block to the following then the tests still pass for me:

```python
    validator = VALIDATORS.get(record_type)
    if validator:
        result = validator(datum, schema=schema, raise_errors=raise_errors)
```

Is there a testcase that needs to be added to show why the extra checking is necessary?

fastavro/_write_py.py
```diff
@@ -296,89 +312,174 @@ def write_map(fo, datum, schema):
...
+                                    field=None,
+                                    raise_errors=raise_errors)
+        else:
+            return False
+
+    errors = []
+    for s in schema:
+        try:
+            ret = validate(datum, schema=s,
+                           field=None,
```
scottbelden: I think this causes the eventual exception (if raised) to have a field value of `None`. Is this intended?

fastavro/_write_common.py
```diff
@@ -0,0 +1,18 @@
+
+class ValidationException(Exception):
+    def __init__(self, datum, schema, field):
+        message = '{field} is {datum} of type ' \
```
scottbelden: We might want to omit the `{field} is` if `field` is `None`. When messing around I got an error like `None is None of type <NoneType> expected str` which was initially a little confusing.

scottbelden:
@rhaarm If you check the box to allow commits from maintainers I can make some updates to help get this through. Thanks!

rhaarm:
Strange if it won't let you, since it's already set to that.
image

I'm going to add some comments, I have a lot of agreement, there were many things I thought of after the final commit. Especially around the Exception/Error handling.

scottbelden:
Ah, I apologize. I didn't actually try to push and commits.

fastavro/_write_common.py
rhaarm: I think I like the first, it might be possible to make things lighter weight too with slots or namedtuple class for the ValidationError object.

fastavro/_write_py.py
rhaarm: Ya when I saw the `UnknownType` in schema, I was assuming it was meant for here but never migrated for use.

fastavro/_write_py.py
rhaarm: As mentioned, lets go with the previous option

fastavro/_write_py.py
rhaarm: Ah because this is only joining the `namespace.name`, not the `namespace.name.type`, should be a simple fix.

fastavro/_write_py.py
rhaarm: I was having failed test previously, the "null" check needs to happen before the `if datum is None:` check. If the "null" or "union with null" is possible, then the None datum check should not occur. Maybe I need to update test_validation.py to verify this.

fastavro/_write_common.py
rhaarm: Ya I was going for simple approach to start, so ya maybe when updating ValidationError object, it can customize the message for None.

fastavro/write.py
```diff
@@ -13,8 +14,11 @@
 
 WRITERS = _write.WRITERS
 LOGICAL_WRITERS = _write.LOGICAL_WRITERS
+VALIDATORS = _write.VALIDATORS
+validate = _write.validate
```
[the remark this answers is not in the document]
rhaarm: I agree, I felt this needed to be moved from the get go, but wanted to keep it in the context that it was already. So I think we should rip it out for this version.

rhaarm:
I'm going to take a stab at updating this code. I just did a rebase with upstream/master, now I'll split the validation to it's own files and referencing with write files. Also, apply the changes that were suggested.

scottbelden:
Sounds good. Thanks! Let me know if you run into any issues along the way.

fastavro/_write_common.py
rhaarm: Single error.
```
  File "/app/fastavro/fastavro/_validate_py.py", line 175, in validate
    raise ValidationErrors(ValidationError(datum, schema, ns_field))
fastavro._validate_common.ValidationErrors: [
  "namespace.missingerror.str is 11 of type <class 'int'> expected string"
]
```

Multiple errors on same union field.
```
  File "/app/fastavro/fastavro/_validate_py.py", line 126, in validate_union
    raise ValidationErrors(*errors)
fastavro._validate_common.ValidationErrors: [
  "namespace.missingerror.str_null is 11 of type <class 'int'> expected null",
  "namespace.missingerror.str_null is 11 of type <class 'int'> expected string"
]
```

rhaarm: Should ValidationErrors be changed to ValidationException?

scottbelden: Ugh, I'm kind of torn... I would say to keep `ValidationErrors` as it's more similar to normal types of exceptions (e.g. `TypeError`, `ValueError`, `KeyError`, `OSError`, etc), but I wouldn't want someone to make a typo by omitting the final `s`.

I guess I'm slightly still in favor of `ValidationErrors` but could be convinced otherwise.

fastavro/write.py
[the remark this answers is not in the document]
rhaarm: I was running benchmarks and had a freak out moment that it was running slower! But I realized that the "_validate.c" didn't get created...
Added to the `setup.py`...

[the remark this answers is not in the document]
scottbelden: You might have to make some changes to the Makefile as well.

fastavro/_write_py.py
rhaarm: I have a fix for the "breaking change", we update UnknownType to inherit `ValueError`.

```
class UnknownType(ValueError):
```

I've made the change and tested in main, would you like me to create any unittest around it?

rhaarm:
Look like AppVeyor build broke on install of flake8 :(
How do I kick off another test? Do a minor commit? I guess I should put in some docs...

scottbelden:
Lame. I kicked it off again.

scottbelden:
Thanks. I took a brief look at this earlier and I know there were a few things I had comments/questions about so I'll try to find some time tomorrow or in the coming days to give this a full review.

rhaarm:
Pushing new updates soon, I had some extra ideas that I applied as well around adding new validators. I also, set it up to not all the basic validators be overridden, what are your thoughts around that? Should we allow for updating VALIDATORS, so that "validate_fixed" can be overridden for custom logicalTypes, like possibly MD5?

fastavro/_write_py.py
```diff
@@ -13,18 +13,16 @@
 import time
 import uuid
 from binascii import crc32
-from collections import Iterable, Mapping
 from os import urandom, SEEK_SET
 from struct import pack
 from zlib import compress
-import numbers
 
+from fastavro._validate_py import validate
```
scottbelden: I think this can be `from .validate import validate` to match how the others are done.

fastavro/_write.pyx
```diff
@@ -5,24 +5,19 @@
...
-from libc.string cimport memset
-from os import urandom, SEEK_SET
+from os import urandom
 from zlib import compress
-import numbers
 
 from fastavro import const
-from .const import DAYS_SHIFT
-from ._six import utob, long, is_str, iterkeys, itervalues, iteritems, mk_bits
+from .validate import validate
```
scottbelden: It shouldn't technically matter, but we should change this to `from ._validate import validate` to match the other imports.

fastavro/validate.py
```diff
@@ -0,0 +1,42 @@
...
+    writer(out, schema, records)
+'''
+try:
+    from . import _validate
+except ImportError:
+    from . import _validate_py as _validate
+from ._validate_common import ValidationErrorData, ValidationError
+
+validate = _validate.validate
+registry_validator = _validate.register_validator
```
scottbelden: I like the name `register_validator` more than `registry_validator`

benchmark/benchmark.py
```diff
@@ -3,6 +3,23 @@
 import time
 
 from fastavro import writer, reader
+try:
+    from datetime.timezone import utc
+except ImportError:
+    from fastavro._timezone import utc
+
+try:
+    from fastavro.validate import validate, validate_many
```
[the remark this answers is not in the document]
rhaarm: I did this, so that I could test benchmark between versions.

fastavro/_schema.pyx
```diff
@@ -28,7 +28,7 @@ cpdef inline str extract_logical_type(schema):
     return None
 
 
-def schema_name(schema, parent_ns):
+cpdef schema_name(object schema, parent_ns):
```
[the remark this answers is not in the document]
rhaarm: Was trying to see if it would speed things up, from what I saw it didn't introduce any bugs, but this can be reject or reverted if you're worried about it.

fastavro/_validate_common.py
```diff
@@ -0,0 +1,39 @@
+from collections import namedtuple
+
+try:
+    import ujson as json
```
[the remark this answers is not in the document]
rhaarm: cool

fastavro/_validate_common.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, *errors):
+        # message = ', '.join(str(e) for e in errors)
+        message = json.dumps([str(e) for e in errors],
+                             indent=2,
+                             ensure_ascii=False)
+        super(ValidationError, self).__init__(message)
+        self.errors = errors
+
+
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
rhaarm: Meant to remove this, I was playing with the formatting of the string. If this moves to test_validation.py, should we test to make sure the string output matches expectation?

fastavro/_validate_py.py
```diff
@@ -0,0 +1,357 @@
...
+    elif field:
+        ns_field = field
+
+    validator = get_validator(record_type)
+    if validator:
+        result = validator(datum, schema=schema,
+                           parent_ns=ns_field,
+                           raise_errors=raise_errors)
+
+    if record_type in SCHEMA_DEFS and result is None:
```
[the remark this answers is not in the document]
rhaarm: Yup totally more valid

fastavro/_write_py.py
rhaarm: I believe this was from when I was working on the Python version only.

fastavro/validate.py
rhaarm: I think this is a typo because I'm also working on Schema Registry things right now.

fastavro/_validate_py.py
```diff
@@ -0,0 +1,357 @@
...
+    'error': validate_record,
+    'request': validate_record
+}
+
+VALIDATORS = BASE_VALIDATORS.copy()
+
+
+def register_validator(record_type, validator):
+    if record_type in BASE_VALIDATORS:
+        raise ValueError("Not allowed to override Base Validators.")
```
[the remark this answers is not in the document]
rhaarm: The only reasoning I could think of, is if we wanted to support numpy types for int,long, it would easy for someone to just register on top of it until we get an actual pull request to update the original.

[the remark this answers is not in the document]
rhaarm: Technically updates to any types in BASE_VALIDATORS will only work for the python version, since the validate function calls the functions directly for speed, if we want to lose the speed, we can have the `cpdef validate` use `get_validator` instead of the large if conditions.

fastavro/_schema.pyx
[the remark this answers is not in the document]
scottbelden: I noticed you opened an issue on this, so let's discuss there and potentially make changes in another PR.

rhaarm:
Yup, I made comment about this subject to move it to another Issue/PR.

rhaarm:
Mostly approved, just need the `benchmark/benchmark.py` updated to use `fastavro.validation`.

scottbelden:
@rhaarm Thanks for all the work towards this. It will be released shortly in 0.19.4.

Issue #112, opened 2017-11-12
Integration with pandas


yiga2:
Now that fasparquet is integrated with pandas, any talk/plan to extend to fastavro ?

tebeka:
We don't have plan for that. You mean to have `DataFrame.to_avro`? That's up to the Pandas team, we'll gladly help.

barrywhart:
@yiga2: Can you provide any details, documentation, examples, etc. on the fastparquet-pandas integration?

This issue may overlap somewhat with #76, which requests allowing fastavro's writer to accept records in list or tuple/namedtuple format (in addition to the currently supported dictionary format).

barrywhart:
Section 5.2.1 of [this PDF](https://media.readthedocs.org/pdf/fastparquet/latest/fastparquet.pdf) has some documentation. If I read this correctly,  it seems that the integration is in the fastparquet library, not the pandas library.

barrywhart:
Source code for `to_pandas()` is [here](https://github.com/dask/fastparquet/blob/master/fastparquet/api.py#L350).

barrywhart:
Both fastparquet and the Apache Avro project from which fastavro was forked both appear to use the same Apache 2.0 license, so it seems like fastavro could reuse code from fastparquet's Pandas integration, if it seemed useful to do so.

NightFantomJ2:
Both numpy's integral types (np.int64 etc) and python's give true on `isinstance(num, numbers.Integral)`
[https://docs.python.org/3/library/numbers.html](https://docs.python.org/3/library/numbers.html), 

changing the isinstance lines for int and long here:
https://github.com/tebeka/fastavro/blob/8b2d2e533da9d890978380f83a04eaa41d6a7527/fastavro/_write_py.py#L330-L344
and here:
https://github.com/tebeka/fastavro/blob/8b2d2e533da9d890978380f83a04eaa41d6a7527/fastavro/_write.pyx#L429-L443

should fix all integral numpy integration.
(works locally without any other changes)
Analogous fixes could be done for floating point types and so forth I presume.

Any reasons not to? I can write up a pull request if you want.

scottbelden:
I'll be honest; I don't use `pandas` at all and `numpy` all that much, so I've never really understood what the goal of this issue is. However, if it can be resolved by making the change to `numbers.Integral` that you mentioned, then that sounds good to me.

Feel free to submit the PR. Thanks!

barrywhart:
[...]

PR #183, opened 2018-05-09
added abstract base classes to integral and float types

NightFantomJ2:
added abstract base classes to integral and float types in _write.validate, to support numpy types (and other classes that correctly register the abstract base classes), including tests

fixes part of https://github.com/tebeka/fastavro/issues/112, as per discussion there

### DIFF
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -18,6 +18,7 @@ from cpython.tuple cimport PyTuple_GET_ITEM
 from libc.string cimport memset
 from os import urandom, SEEK_SET
 from zlib import compress
+import numbers
 
 from fastavro import const
 from .const import DAYS_SHIFT
@@ -428,7 +429,7 @@ cpdef validate(object datum, schema):
 
     if record_type == 'int':
         return (
-            (isinstance(datum, (int, long,)) and
+            (isinstance(datum, (int, long, numbers.Integral)) and
              INT_MIN_VALUE <= datum <= INT_MAX_VALUE) or
             isinstance(datum, (
                 datetime.time, datetime.datetime, datetime.date))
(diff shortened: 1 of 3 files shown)


NightFantomJ2:
For those checks to succeed I probably need to add numpy somewhere in the config files? Can someone help me with that? Alternatively we can disable the numpy test and the other test will still confirm that all still works as intended.

I don't understand how pypy errors while the others don't
`AttributeError: module 'fastavro' has no attribute '_write'`
Should I access these in a different way? Or shouldn't I call validate like that?

scottbelden:
To add numpy you should just need to add it to the [travis file](https://github.com/tebeka/fastavro/blob/master/.travis.yml#L11) and the [appveyor file](https://github.com/tebeka/fastavro/blob/master/appveyor.yml#L15). I created an issue just now so that when I find some time I can make this a little easier: https://github.com/tebeka/fastavro/issues/184.

For `validate`, you will need to add it to `write.py` like what is being done in https://github.com/tebeka/fastavro/pull/182: https://github.com/tebeka/fastavro/pull/182/files#diff-b775c6f9a6a096017db8a44fc0aad380R18. However, for the moment, I think it would be best to have it in that file as `_validate = _write.validate`. `validate` isn't really a public API and while it will become so in the other PR, I think it might make it's way into it's own module so let's leave an underscore until that gets figured out.

After that you should be able to change the test to do `fastavro.write._validate`.

scottbelden:
Oh, and just for extra knowledge, the reason why you get `AttributeError: module 'fastavro' has no attribute '_write'` on pypy is because `fastavro._write` refers to this file (https://github.com/tebeka/fastavro/blob/master/fastavro/_write.pyx) which isn't compiled or accessible on pypy. It can get a little messy, but all the public functions should be accessed from [`fastavro.write`](https://github.com/tebeka/fastavro/blob/master/fastavro/write.py#L1-L4) which does an import check to use the cython files if they exist but fall back to pure python if they don't (which is the case for pypy).

NightFantomJ2:
I'll see if I can do those fixes tonight.

Thanks for the clarifications, @scottbelden!

scottbelden:
@NightFantomJ2 If you check the box to allow commits from maintainers I can make some updates to help get this through. Thanks!

scottbelden:
Actually, I went ahead and pushed a new branch with your original commit and a commit to fix the issues. The PR is here: https://github.com/tebeka/fastavro/pull/189.

PR #188, opened 2018-05-14
add json reader/writer

scottbelden:
Resolves https://github.com/tebeka/fastavro/issues/33

### DIFF
--- a/fastavro/json_writer.py
+++ b/fastavro/json_writer.py
@@ -3,7 +3,7 @@ import json
 from .schema import extract_record_type
 from .six import iteritems, btou
 from .write import acquaint_schema
-from ._write_py import validate
+from .write import _validate as validate
 
 
 def _write_json(datum, schema):
--- a/fastavro/json_writer.py
+++ b/fastavro/json_writer.py
@@ -1,0 +1,131 @@
+import json
+
+from .schema import extract_record_type
+from .six import iteritems
+from .write import acquaint_schema
+from ._write_py import validate
+
+
+def _write_json(datum, schema):
+    record_type = extract_record_type(schema)
+
+    passthrough_schemas = (
+        'null',
+        'boolean',
+        'string',
+        'int',
+        'long',
+        'float',
+        'double',
+        'enum',
+    )
+
+    if record_type in passthrough_schemas:
+        return datum
+
+    elif record_type == 'bytes':
+        return datum.decode('iso-8859-1')
+
+    elif record_type == 'fixed':
+        return datum.decode('iso-8859-1')
+
+    elif record_type == 'union':
+        best_match_index = -1
+        most_fields = -1
+        for index, candidate in enumerate(schema):
+            if validate(datum, candidate):
+                if extract_record_type(candidate) == 'record':
+                    fields = len(candidate['fields'])
+                    if fields > most_fields:
+                        best_match_index = index
+                        most_fields = fields
+                else:
+                    best_match_index = index
+                    break
+        if best_match_index < 0:
+            pytype = type(datum)
+            msg = '%r (type %s) do not match %s' % (datum, pytype, schema)
+            raise ValueError(msg)
+
+        best_match_schema = schema[best_match_index]
+
+        if best_match_schema == 'null':
+            return None
+        else:
+            best_records_type = extract_record_type(best_match_schema)
+            if best_records_type in ('record, enum, fixed'):
+                key = best_match_schema['name']
+            else:
+                key = best_match_schema
+            return {key: _write_json(datum, best_match_schema)}
+
+    elif record_type == 'array':
+        dtype = schema['items']
+        return [_write_json(item, dtype) for item in datum]
+
+    elif record_type == 'map':
+        vtype = schema['values']
+        result = {}
+        for key, value in iteritems(datum):
+            result[key] = _write_json(value, vtype)
+        return result
+
+    elif record_type in ('record', 'error', 'request',):
+        result = {}
+        for field in schema['fields']:
+            result[field['name']] = _write_json(
+                datum.get(field['name'], field.get('default')),
+                field['type'],
+            )
+        return result
+
+    else:
+        raise ValueError('unknown record type - %s' % record_type)
+
+
+def writer(fo,
+           schema,
+           records):
+    """Write records in json form to fo (stream) according to schema
+
+    Parameters
+    ----------
+    fo: file-like
+        Output stream
+    schema: dict
+        Schema of the records
+    records: iterable
+        Records to write
+
+
+
+    Example::
+
+        from fastavro.json_writer import writer
+
+        schema = {
+            'doc': 'A weather reading.',
+            'name': 'Weather',
+            'namespace': 'test',
+            'type': 'record',
+            'fields': [
+                {'name': 'station', 'type': 'string'},
+                {'name': 'time', 'type': 'long'},
+                {'name': 'temp', 'type': 'int'},
+            ],
+        }
+
+        records = [
+            {u'station': u'011990-99999', u'temp': 0, u'time': 1433269388},
+            {u'station': u'011990-99999', u'temp': 22, u'time': 1433270389},
+            {u'station': u'011990-99999', u'temp': -11, u'time': 1433273379},
+            {u'station': u'012650-99999', u'temp': 111, u'time': 1433275478},
+        ]
+
+        with open('weather.json', 'w') as out:
+            writer(out, schema, records)
+    """
+    acquaint_schema(schema)
+    for record in records:
+        json.dump(_write_json(record, schema), fo)
+        fo.write('\n')
--- a/fastavro/json_writer.py
+++ b/fastavro/json_writer.py
@@ -1,7 +1,7 @@
 import json
 
 from .schema import extract_record_type
-from .six import iteritems
+from .six import iteritems, btou
 from .write import acquaint_schema
 from ._write_py import validate
 
@@ -24,10 +24,10 @@ def _write_json(datum, schema):
         return datum
 
     elif record_type == 'bytes':
-        return datum.decode('iso-8859-1')
+        return btou(datum, encoding='iso-8859-1')
 
     elif record_type == 'fixed':
-        return datum.decode('iso-8859-1')
+        return btou(datum, encoding='iso-8859-1')
 
     elif record_type == 'union':
         best_match_index = -1
--- a/fastavro/json_writer.py
+++ b/fastavro/json_writer.py
@@ -3,7 +3,7 @@ import json
 from .schema import extract_record_type
 from .six import iteritems, btou
 from .write import acquaint_schema
-from .write import _validate as validate
+from .validation import validate
 
 
 def _write_json(datum, schema):
@@ -33,7 +33,7 @@ def _write_json(datum, schema):
         best_match_index = -1
         most_fields = -1
         for index, candidate in enumerate(schema):
-            if validate(datum, candidate):
+            if validate(datum, candidate, raise_errors=False):
                 if extract_record_type(candidate) == 'record':
                     fields = len(candidate['fields'])
                     if fields > most_fields:
--- a/fastavro/json_writer.py
+++ b/fastavro/json_writer.py
@@ -1,8 +1,7 @@
 import json
 
-from .schema import extract_record_type
+from .schema import extract_record_type, parse_schema
 from .six import iteritems, btou
-from .write import acquaint_schema
 from .validation import validate
 
 
@@ -125,7 +124,7 @@ def writer(fo,
         with open('weather.json', 'w') as out:
             writer(out, schema, records)
     """
-    acquaint_schema(schema)
+    parsed_schema = parse_schema(schema, _write_hint=False)
     for record in records:
-        json.dump(_write_json(record, schema), fo)
+        json.dump(_write_json(record, parsed_schema), fo)
         fo.write('\n')
--- a/fastavro/json_writer.py
+++ b/fastavro/json_writer.py
@@ -52,7 +52,7 @@ def _write_json(datum, schema):
             return None
         else:
             best_records_type = extract_record_type(best_match_schema)
-            if best_records_type in ('record, enum, fixed'):
+            if best_records_type in ('record', 'enum', 'fixed'):
                 key = best_match_schema['name']
             else:
                 key = best_match_schema
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -1,6 +1,7 @@
 from fastavro import json_writer, json_reader
 from fastavro.six import StringIO
 
+import json
 import pytest
 
 pytestmark = pytest.mark.usefixtures("clean_readers_writers_and_schemas")
@@ -184,6 +185,79 @@ def test_more_than_one_record():
     }, {
         'string': 'bar',
         'int': 2,
+    }]
+
+    new_records = roundtrip(schema, records)
+    assert records == new_records
+
+
+def test_encoded_union_output():
+    schema = {
+        "type": "record",
+        "name": "Test",
+        "namespace": "test",
+        "fields": [{
+            "name": "union",
+            "type": [
+                'null',
+                'int',
+                {
+                    "type": "record",
+                    "name": "union_record",
+                    "fields": [{
+                        "name": "union_record_field",
+                        "type": "string",
+                    }],
+                },
+            ]
+        }]
+    }
+
+    # A null value is encoded as just null
+    records = [{'union': None}]
+    new_file = StringIO()
+    json_writer.writer(new_file, schema, records)
+    assert new_file.getvalue().strip() == json.dumps({"union": None})
+
+    # A non-null, non-named type is encoded as an object with a key for the
+    # type
+    records = [{'union': 321}]
+    new_file = StringIO()
+    json_writer.writer(new_file, schema, records)
+    assert new_file.getvalue().strip() == json.dumps({"union": {'int': 321}})
+
+    # A non-null, named type is encoded as an object with a key for the name
+    records = [{'union': {'union_record_field': 'union_field'}}]
+    new_file = StringIO()
+    json_writer.writer(new_file, schema, records)
+    expected = json.dumps({
+        "union": {
+            'union_record': {
+                'union_record_field': 'union_field'
+            }
+        }
+    })
+    assert new_file.getvalue().strip() == expected
+
+
+def test_union_string_and_bytes():
+    schema = {
+        "type": "record",
+        "name": "Test",
+        "namespace": "test",
+        "fields": [{
+            "name": "union",
+            "type": [
+                'string',
+                'bytes',
+            ]
+        }]
+    }
+
+    records = [{
+        'union': 'asdf',
+    }, {
+        'union': b'asdf'
     }]
 
     new_records = roundtrip(schema, records)
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -1,5 +1,5 @@
 from fastavro import json_writer, json_reader
-from fastavro.six import MemoryIO
+from fastavro.six import StringIO
 
 import pytest
 
@@ -7,7 +7,7 @@ pytestmark = pytest.mark.usefixtures("clean_readers_writers_and_schemas")
 
 
 def roundtrip(schema, records):
-    new_file = MemoryIO()
+    new_file = StringIO()
     json_writer.writer(new_file, schema, records)
     new_file.seek(0)
 
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -1,0 +1,190 @@
+from fastavro import json_writer, json_reader
+from io import StringIO
+
+import pytest
+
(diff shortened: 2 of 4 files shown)


scottbelden:
Not ready for merge. Just wanted to put this here so CI could test it.

tests/test_json.py
```diff
@@ -0,0 +1,190 @@
...
+            "type": {
+                "type": "fixed",
+                "name": "fixed_field",
+                "size": 5,
+            },
+        }, {
+            "name": "union",
+            "type": [
+                'null',
+                'int',
```
mvallebr: I would add more types here and test cases where union type is enum, fixed, float, double, etc. 
As many avro types have more than 1 correspondent in python, the complete implementation should handle more corner cases, like defining that the library should accept `decimal` as python correspondent for `double`, for instance. 

What would you do, for example, if you had 2 types in this union: string and bytes ? Both correspond to python strings.

I think basic cases like this are already worth merging, but they aren't the full thing. It would be possible to support something even more complete than current python avro implementation.

tests/test_json.py
```diff
@@ -0,0 +1,190 @@
...
+        'boolean': True,
+        'string': 'foo',
+        'bytes': b'\xe2\x99\xa5',
+        'int': 1,
+        'long': 1 << 33,
+        'float': 2.2,
+        'double': 3.3,
+        'fixed': b'\x61\x61\x61\x61\x61',
+        'union': {
+            'union_record_field': 'union_field',
```
mvallebr: This is wrong. The correct datum should be:
```
'union': {  'union_record': {'union_record_field': 'union_field' }}
```

tests/test_json.py
```diff
@@ -0,0 +1,190 @@
...
+        'null': None,
+        'boolean': True,
+        'string': 'foo',
+        'bytes': b'\xe2\x99\xa5',
+        'int': 1,
+        'long': 1 << 33,
+        'float': 2.2,
+        'double': 3.3,
+        'fixed': b'\x61\x61\x61\x61\x61',
+        'union': 321,
```
mvallebr: This is wrong. The correct datum would be:
```
'union': {  'int': 321}
```

tests/test_json.py
```diff
@@ -0,0 +1,190 @@
...
+        'null': None,
+        'boolean': True,
+        'string': 'foo',
+        'bytes': b'\xe2\x99\xa5',
+        'int': 1,
+        'long': 1 << 33,
+        'float': 2.2,
+        'double': 3.3,
+        'fixed': b'\x61\x61\x61\x61\x61',
+        'union': None,
```
mvallebr: This is right. When value is `null`, there is no type specification.

tests/test_json.py
scottbelden: I agree. I've been thinking about this recently and it would be really nice to have some helper (or maybe a standalone library) that given a schema it would produce 1) a valid datum for that schema using dummy values and 2) be able to create all the different types of valid datum (so in the case of unions, create multiple dictionaries where each is a different union case).

Without that existing, I think as you mentioned there should be some more common test cases. I just haven't added those yet as I wanted to just see if things were on the right track first. I'll add a comment to my top comment to say what needs to be done.

fastavro/json_writer.py
```diff
@@ -0,0 +1,131 @@
...
+                        most_fields = fields
+                else:
+                    best_match_index = index
+                    break
+        if best_match_index < 0:
+            pytype = type(datum)
+            msg = '%r (type %s) do not match %s' % (datum, pytype, schema)
+            raise ValueError(msg)
+
+        best_match_schema = schema[best_match_index]
```
mvallebr: This is exactly the part we want to avoid. In my code, I created a code generator that creates python beans with information about the types they contain. I understand for normal python dictionaries, there is no easy way of detecting which avro type a python type corresponds to, except in really obvious cases. 

This implementation is already worth merging, as it adds a lot of value fastavro currently doesn't support, but I think you could extend the specification here and define how the datum should look like in your implementation to correspond to a certain json, so the result would be deterministic. Of course, this would have to be added to documentations. 

For instance, python only has float native type, avro has double and fixed. Even worse, avro spec supports that an union can have more than 1 fixed type in the union, as long as the precision is different. 

So if fastavro provided an `AvroFixed` class to hold avro fixed items, this problem would be solved. If anyone wants to send a dict to this method, the datum should contains an `AvroFixed` instance to correspond to avro fixed type. Is this a good solution? Not sure. Would `decimal` type be enough instead?

Same for string and bytes blob. If you google it, you will see several users complaining about the official python library because how to represent avro types in python is non-deterministic. We can do better.

tests/test_json.py
mvallebr: Maybe I could write this library, let's just keep discussing first to reach common agreement. I've already spent a lot of time handling many of these complications.

tests/test_json.py
mvallebr: So, the main reason why I asked modification in fastavro itself is - even if I write my wrapper, as I did, to handle the json encoding, when you transform from json encoding to non json encoding you are losing information and that's the part we don't want. 

Unfortunately, I see little value in keeping sending this to `json_writer.writer`.  What we really need is probably to change the c-level library to make it accept json encoding and do the opposite, make the old methods to convert from non-json enconding to json encoding, by making guesses and inferring things.

mvallebr: Actually, thinking on the float/double/fixed use case above, it would be really complicated to send this to C. 
Maybe what should be sent to the C library is neither json spec nor current implementation, is a specific protocol format that both json encoding and normal encoding should be converted to/from.

mvallebr:
Let me share to you a piece of code I wrote in my wrapper to handle schema validation, it's not fully related to this PR, but it relates to the python type vs avro type problem I related above. See how ambiguous things can be for python.

```python

_SCHEMA_TYPE_NATIVE_MATCHER = {
    'null': lambda datum, avro_schema: datum is None,
    'boolean': lambda datum, avro_schema: isinstance(datum, bool),
    'string': lambda datum, avro_schema: isinstance(datum, basestring),
    'bytes': lambda datum, avro_schema: isinstance(datum, str),
    'int': lambda datum, avro_schema: isinstance(datum, int) or isinstance(datum,
                                                                           long) and INT_MIN_VALUE <= datum <= INT_MAX_VALUE,
    'long': lambda datum, avro_schema: isinstance(datum, int) or isinstance(datum,
                                                                            long) and LONG_MIN_VALUE <= datum <= LONG_MAX_VALUE,
    'float': lambda datum, avro_schema: isinstance(datum, int) or isinstance(datum, long) or isinstance(datum, float),
    'double': lambda datum, avro_schema: isinstance(datum, int) or isinstance(datum, long) or isinstance(datum, float),
    'fixed': lambda datum, avro_schema: isinstance(datum, str) and len(datum) == avro_schema.size,
    'enum': lambda datum, avro_schema: datum in avro_schema.symbols,
    'array': lambda datum, avro_schema: isinstance(datum, list) or isinstance(datum, set),
    'map': lambda datum, avro_schema: isinstance(datum, dict),
    'record': lambda datum, avro_schema: isinstance(datum, dict),
    'error': lambda datum, avro_schema: isinstance(datum, dict),
    'request': lambda datum, avro_schema: isinstance(datum, dict),
}

_SCHEMA_TYPE_NATIVE_MATCHER_ERROR_MSG = {
    'null': lambda datum, avro_schema: "{}: {} was supposed to be None, but isn't".format(avro_schema.type, datum),
    'boolean': lambda datum, avro_schema: "{}: {} was supposed to be bool, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'string': lambda datum, avro_schema: "{}: {} was supposed to be basestring, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'bytes': lambda datum, avro_schema: "{}: {} was supposed to be str, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'int': lambda datum,
                  avro_schema: "{}: {} was supposed to be int or long within [{}, {}] range, actual type is {}".format(
        avro_schema.type, datum, INT_MIN_VALUE, INT_MAX_VALUE, type(datum)),
    'long': lambda datum,
                   avro_schema: "{}: {} was supposed to be int or long within [{}, {}] range, actual type is {}".format(
        avro_schema.type, datum, LONG_MIN_VALUE, LONG_MAX_VALUE, type(datum)),
    'float': lambda datum, avro_schema: "{}: {} was supposed to be either int, long or float, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'double': lambda datum, avro_schema: "{}: {} was supposed to be either int, long or float, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'fixed': lambda datum, avro_schema: "{}: {} was supposed to be str with len {}, but is {} with len {}".format(
        avro_schema.type, datum, avro_schema.size, type(datum), len(datum)),
    'enum': lambda datum, avro_schema: "{}: {}/{} was supposed to one of {}".format(
        avro_schema.type, datum, type(datum), avro_schema.symbols),
    'array': lambda datum, avro_schema: "{}: {} was supposed to be list, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'map': lambda datum, avro_schema: "{}: {} was supposed to be dict, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'record': lambda datum,
                     avro_schema: "['record', 'error', 'request']: {}: {} was supposed to be dict, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'error': lambda datum,
                    avro_schema: "['record', 'error', 'request']: {}: {} was supposed to be dict, but is {}".format(
        avro_schema.type, datum, type(datum)),
    'request': lambda datum,
                      avro_schema: "['record', 'error', 'request']: {}: {} was supposed to be dict, but is {}".format(
        avro_schema.type, datum, type(datum)),
}


class SchemaValidateException(Exception):
    pass


def _throw_if_false(condition, message):
    if not condition:
        raise SchemaValidateException(message)


    schema_type = avro_schema.type
    if schema_type in _SCHEMA_TYPE_NATIVE_MATCHER.keys():
        _throw_if_false(_SCHEMA_TYPE_NATIVE_MATCHER[schema_type](datum, avro_schema),
                        _SCHEMA_TYPE_NATIVE_MATCHER_ERROR_MSG[schema_type](datum, avro_schema))

```


And bellow is my full validation function, which tries to handle both json encoding and "normal" encoding for unions, but I think it's a bad idea and I am probably changing that in the future. I created this because the avro default validate function was giving me horrible error messages when the schema wasn't corresponding to the datum and I was spending a lot of time to find the errors in these cases.

```python


def validate(avro_schema, datum, field_name="", permissive=True):
    """
    Determine if a python datum is an instance of a schema.

    :param avro_schema: Schema read with 'read_avro_schema'
    :param datum: dict corresponding to a python object to be validated
    :param field_name: when validating fields of records, the field currently being validated is passed as an argument,
    to enable the inner level of recursion to generate a proper error message when needed.
    :param permissive: if set, gens warnings when input is not avro json encoded. Otherwise, SchemaValidateException.
    :return: None if successful. Raises SchemaValidateException on error.
    """

    def validate_sub_schema(sub_schema, sub_datum, sub_field_name=""):
        try:
            validate(sub_schema, sub_datum, sub_field_name, permissive)
        except SchemaValidateException as e:
            raise SchemaValidateException(
                "{}: {} type {} had errors validating inner field '{}': \n\t{}\n".format(
                    avro_schema.type, datum, type(datum), sub_field_name, str(e)))

    def warn_or_raise(error_msg):
        if permissive:
            logging.warning(error_msg)
        else:
            raise SchemaValidateException(error_msg)

    schema_type = avro_schema.type
    if schema_type in _SCHEMA_TYPE_NATIVE_MATCHER.keys():
        _throw_if_false(_SCHEMA_TYPE_NATIVE_MATCHER[schema_type](datum, avro_schema),
                        _SCHEMA_TYPE_NATIVE_MATCHER_ERROR_MSG[schema_type](datum, avro_schema))

    if schema_type == 'array':
        for d in datum:
            validate_sub_schema(avro_schema.items, d, field_name)

    elif schema_type == 'map':
        for k in datum.keys():
            _throw_if_false(isinstance(k, basestring),
                            "{}: map key {} was supposed to be basestring, but is {}".format(schema_type, k, type(k)))
        for v in datum.values():
            validate_sub_schema(avro_schema.values, v, field_name)

    elif schema_type in ['record', 'error', 'request']:
        for f in avro_schema.fields:
            validate_sub_schema(f.type, datum.get(f.name), f.name)

    elif schema_type in ['union', 'error_union']:
        """ When datum is union, the datum is expected to be something like
        {"ClassName": {...}} where {...} is the datum corresponding to the "ClassName" schema
        Therefore, the dict should have just 1 key.
        """
        if datum is None:
            matching_schemas = [s for s in avro_schema.schemas if s.type == "null"]
            if len(matching_schemas) == 1:
                return
            else:
                warn_or_raise("Datum {} is null, but no union type is: {}".format(
                    datum, [_schema_name(s) for s in avro_schema.schemas]))
        elif type(datum) == dict and len(datum.keys()) == 1:
            schema_name = datum.keys()[0]
            real_datum = datum[schema_name]
            matching_schemas = [s for s in avro_schema.schemas if
                                _schema_name(s) == schema_name or _schema_fullname(s) == schema_name]
            if len(matching_schemas) == 1:
                validate_sub_schema(matching_schemas[0], real_datum, field_name)
                return
            else:
                warn_or_raise(("Datum {} corresponds to an union object, but it wasn't possible to find schema {} "
                               "in expected list {}").format(datum, schema_name,
                                                             [_schema_name(s) for s in avro_schema.schemas]))

        warn_or_raise("Trying to infer schema type for datum {} from list of expected union types: {}".format(
            datum, [_schema_name(s) for s in avro_schema.schemas]))

        num_valid_schemas = 0
        errors = []
        for s in avro_schema.schemas:
            try:
                validate_sub_schema(s, datum, field_name)
                num_valid_schemas += 1
            except SchemaValidateException as e:
                errors.append(
                    "Datum {} validation for union {} returned error {}".format(str(datum), _schema_fullname(s),
                                                                                str(e)))
        _throw_if_false(num_valid_schemas >= 1,
                        ("{}: Union type has {} valid schemas, but it was supposed to have at least 1\n"
                         "Errors for each union type: \n{}").format(
                            schema_type, num_valid_schemas, "\n".join(errors)))

```

scottbelden:
As for the discussion about validation errors, I agree that the error messages aren't as helpful as they should be. There is an active [PR](https://github.com/tebeka/fastavro/pull/182) that is trying to address that. If you have time, feedback on that PR would be appreciated.

As for type information getting lost, I've seen this as well especially when you have a union of two records either ambiguous differences. While decoding the binary or avro-json you know the type, but unless you preserve that somehow, then if you go to re-encode it you might pick the wrong schema. The same seems to be true with something like a union of a `Fixed` and `bytes` when they are the same length as you mentioned.

I think this is deserving of it's own issue (or probably multiple issues), so I'll begin that. There are a few things we should keep in mind:

1) We should try to remain compatible with the apache `avro` python library in our python representation unless we have an extremely good reason not to. This is important because libraries like [confluent-kafka-python](https://github.com/confluentinc/confluent-kafka-python) try to use `fastavro` and then fall back to `avro`. If for a `fixed` type we return a `Fixed` object and the other library returns `bytes` then this could be very disruptive. If we find a very good reason to differ from their representation then we may need to have some flag that can keep supporting their format.

2) We need to make sure to keep performance in mind. Of course correctness is more important than performance, but it's something that we will want to be aware of.

scottbelden:
Good luck with trying to get anything into the official version. It seems they either only care about the java code or they just don't know python enough to feel comfortable merging anything. Perhaps their original python bindings were written by some intern/developer that has since left?

I'm a little confused on what you mean with "communication between python and C". While we do use cython in order to speed up the reading/writing of the avro binary, no state is held in that cython code. It's also worth noting that on pypy the cython code isn't used at all, so we wouldn't want to do something only in that cython file and not in the normal python file.

I'll try to add a few more test and get this cleaned up some so that it can be merged it. I'll probably document it as somewhat experimental so that if we need to change it drastically then hopefully it shouldn't catch anyone off guard.

mvallebr:
>  Perhaps their original python bindings were written by some intern/developer that has since left?

;-) I got that impression too, hope they don't get offended.  LoL 
That's why I tried fast avro in the first place, just to be clear, if your intention is to have a *complete* python library in the future, I would more than glad try to contribute here instead. What I said was more due my understanding being you don't want to focus on being feature complete, fast is enough. Please tell me whether I am wrong. 

It the later case is true though, I didn't get them why you want to be fully compatible with python implementation rather than with python specification only. This interface for json encoding doesn't have to be compatible with the current one, you could accept a datum with Fixed class instances on this one and keep forcing compatibility with avro impl on the other one. 

> I'm a little confused on what you mean with "communication between python and C". 
I thought `json_writer.writer` was using c-bindings and defining a "Fixed" class, for instance, wouldn't be straightforward.  Maybe I haven't analyzed your full code yet, my bad. 

Side comment: it would be useful to have more contributing guidelines written somewhere. When I tried to run tests on my machine, I wasn't able to easily install all dependencies. A docker file that runs the tests would make the life of contributors much more easy.

scottbelden:
> That's why I tried fast avro in the first place, just to be clear, if your intention is to have a complete python library in the future, I would more than glad try to contribute here instead. What I said was more due my understanding being you don't want to focus on being feature complete, fast is enough. Please tell me whether I am wrong.

I think the original intention of this library might have been to be a stripped down version of the official `avro` with speed in mind. However, over time many features (such as schema migration, logical types, schemaless operations, etc) have been added to where at this point I believe this library is not only faster than the official package but now has more features as well.

> It the later case is true though, I didn't get them why you want to be fully compatible with python implementation rather than with python specification only.

From my personal experience, I started with using the official `avro` library and found `fastavro` later. The libraries do have different interfaces, and so when I moved from `avro` to `fastavro` I had to tweak how I did the reading/writing calls, but one convenience was that dictionary on the inside was the same. That made it relatively simple to make the switch.

If we can include the type information (to solve issues like [the ambiguous types](https://github.com/tebeka/fastavro/issues/192)) and still keep it easy for people to switch, I think that would be ideal. Perhaps that can be done by doing things like the following:

```python
class fixed(bytes):
    ...
```

And so when we read the binary we can create a `fixed` type to distinguish from `bytes`. We wouldn't want to force people to use this type (as it would be a pretty massive breaking change and it's only really needed in the case of unions) so we'd either need some `strict` flag that one could opt in to or have some logic to accept both types but prefer the correct one.

I think this could be a good discussion in its own issue.

> Side comment: it would be useful to have more contributing guidelines written somewhere. When I tried to run tests on my machine, I wasn't able to easily install all dependencies. A docker file that runs the tests would make the life of contributors much more easy.

Thanks for the feedback. I recently added a `developer_requirements.txt` to make it more clear about the requirements, but you are correct that the README probably needs some updating. Also, I'll try to add a dockerfile. I don't think I'll use it, but I could see it being useful for others and it should just take a few minutes.

scottbelden:
An example Dockerfile was also posted by @mvallebr at https://github.com/mvallebr/fastavro/blob/fb0319c28ee1a275e1316bee372bc59a0c5d3e35/Dockerfile

I was just experimenting with a modified version of that that looks like the following:

```
FROM debian:latest

RUN apt-get update &&  apt-get install -y --allow-downgrades --allow-remove-essential --allow-change-held-packages \
    python3 python3-pip python2.7 python-pip python3.4 python3.5 python 3.6 git

COPY . /source
WORKDIR /source

RUN pip install virtualenv

RUN virtualenv -p python2.7 testenv \
    && . testenv/bin/activate \
    && pip install -r /source/developer_requirements.txt \
    && ./run-tests.sh

RUN virtualenv -p python3.5 testenv3 \
    && . testenv3/bin/activate \
    && pip install -r /source/developer_requirements.txt \
    && ./run-tests.sh
```

This would run both the 2.7 and 3.5 tests.

mvallebr:
Guys, this seems brilliant. Let me process all this info and I will get back to you soon

scottbelden:
@mvallebr If you have more thoughts/opinions on the dockerfile please add them over at https://github.com/tebeka/fastavro/pull/193.

mvallebr:
Guys, sorry for the delay to answer. I will take about 2 weeks to be able to look deeply into this again, but I will definitely get back to you regarding every point. 

Sorry about the delay!

scottbelden:
No worries. Overall I'm not sure how I feel about this in general. I haven't put in the logical type support, and when I do I think it will expose a problem. That being that this will be very heavily duplicated from the normal `writer` and `reader`. I think there should be a better way to avoid this duplication.

fastavro/json_writer.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if best_match_schema == 'null':
+            return None
+        else:
+            best_records_type = extract_record_type(best_match_schema)
+            if best_records_type in ('record, enum, fixed'):
```
Artimi:
```suggestion
            if best_records_type in ('record', 'enum', 'fixed'):
```
I believe this was the intent.

scottbelden: Yes, definitely. I should put in a test that would have caught this.

scottbelden:
Implemented in https://github.com/fastavro/fastavro/pull/343

PR #193, opened 2018-05-15
add dockerfile for easy testing

scottbelden:


rhaarm:
Wow, this takes a long time to build!

Also, for those coming from Windows, I've made improvements. It's stupid I have to do the carriage return sed in the container, but ya, it's a thing... Also, I had errors when trying to find `./run-tests.sh`, so I was explicit on the path and using `/bin/bash`.

```
FROM debian:latest

RUN apt-get update &&  apt-get install -y --allow-downgrades --allow-remove-essential --allow-change-held-packages \
    python3 python3-pip python2.7 python-pip python3.5 python 3.6 git

RUN pip install virtualenv

COPY . /source
WORKDIR /source
RUN sed -i -e 's/\r//' /source/run-tests.sh

RUN virtualenv -p python2.7 testenv \
    && . testenv/bin/activate \
    && pip install -r /source/developer_requirements.txt \
    && /bin/bash /source/run-tests.sh

RUN virtualenv -p python3.5 testenv3 \
    && . testenv3/bin/activate \
    && pip install -r /source/developer_requirements.txt \
    && /bin/bash /source/run-tests.sh
```

PR #207, opened 2018-05-27
fix overzealous validation errors

ryan-williams:
Before this change, benchmarking spuriously fails:

```bash
./run-tests.sh
python benchmark/benchmark.py
```
```
Small - 1 rec - 100,000 runs
Writing 1 records to one file...
... 100000 runs averaged 2.65738511086e-05 seconds
Reading 1 records from one file...
... 100000 runs averaged 2.72266125679e-05 seconds
Validating 1 records from one file...
  File "benchmark/benchmark.py", line 161, in <module>
    valid = validater(schema, original_records, runs=num_runs)
  File "benchmark/benchmark.py", line 28, in validater
    valid = validate_many(records, schema)
  File "fastavro/_validation.pyx", line 228, in fastavro._validation.validate_many
    cpdef validate_many(records, schema, bint raise_errors=True):
  File "fastavro/_validation.pyx", line 239, in fastavro._validation.validate_many
    raise ValidationError(*errors)
fastavro._validate_common.ValidationError: []
```

`validate_many` is the culprit, though mysteriously `validate_union` seems to have the same issue (fixed here), but is apparently never exercised by any tests?

### DIFF
--- a/fastavro/_validation.pyx
+++ b/fastavro/_validation.pyx
@@ -155,7 +155,7 @@ cdef inline bint validate_union(object datum, list schema, str parent_ns=None,
                 return True
         except ValidationError as e:
             errors.extend(e.errors)
-    if raise_errors and errors:
+    if raise_errors:
         raise ValidationError(*errors)
     return False
 
--- a/fastavro/_validation.pyx
+++ b/fastavro/_validation.pyx
@@ -155,7 +155,7 @@ cdef inline bint validate_union(object datum, list schema, str parent_ns=None,
                 return True
         except ValidationError as e:
             errors.extend(e.errors)
-    if raise_errors:
+    if raise_errors and errors:
         raise ValidationError(*errors)
     return False
 
@@ -235,6 +235,6 @@ cpdef validate_many(records, schema, bint raise_errors=True):
             results.append(result)
         except ValidationError as e:
             errors.extend(e.errors)
-    if raise_errors:
+    if raise_errors and errors:
         raise ValidationError(*errors)
     return all(results)
--- a/fastavro/_validation_py.py
+++ b/fastavro/_validation_py.py
@@ -412,6 +412,6 @@ def validate_many(records, schema, raise_errors=True):
             results.append(validate(record, schema, raise_errors=raise_errors))
         except ValidationError as e:
             errors.extend(e.errors)
-    if raise_errors:
+    if raise_errors and errors:
         raise ValidationError(*errors)
     return all(results)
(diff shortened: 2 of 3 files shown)


rhaarm:
The new validation was tested with unittest and benchmark. If something fails to validate true it raises the error when set true to raise errors. So if its returning false with an empty errors list then we need to fix the underlying bug around raise errors.  Can you provide the data/schema that's causing the error?

ryan-williams:
I'm confused / think we're miscommunicating. AFAICT:

* since https://github.com/tebeka/fastavro/commit/36a0f3dfe312d3cf00f810a3f909a9b2ae799042 6d ago, `validate_many` and `validate_union` [will incorrectly always throw a `ValidationError`, even when no errors were found](https://github.com/ryan-williams/fastavro/blob/36a0f3dfe312d3cf00f810a3f909a9b2ae799042/fastavro/_validation.pyx#L375-L376), unless `raise_errors` is explicitly passed as `False`
  * this situation would have arisen previously anytime `raise_errors` was passed as `True`, but the default was `False` (and that was used at all call-sites?)
* `python benchmark/benchmarks.py` hits this bug and fails with a `ValidationError` where there should have been none
  * this script is not run in any of the CIs, so it was not noticed as broken
* this change fixes things so that a `ValidationError` is only thrown when validation errors occur

lmk if I'm missing something or that doesn't make sense, thanks

rhaarm:
Wow, totally correct. Possiy Scott can get this merged. Can you add a unit test to prove that raise does not raise when true?

scottbelden:
Good catch. I changed `validate` to have `raise_errors` True by default (with the thought that later we might have another `is_valid` that could set `raise_errors` to False as a boolean check.

as @rhaarm mentioned, it would be good to have a test for this. I think it should be sufficient to modify this test: https://github.com/tebeka/fastavro/blob/19f7c6e30f6e225652a12e4cea35f1b88ef3fcec/tests/test_validation.py#L66-L72

And simply add the following below the assert:

```python
# Should not raise any exceptions
validation_raise(schema, *records)
```

When I make that change locally without your fix I get a failure which exposes the problem.

After making this change it will expose that we need to make the same change in `_validation.pyx` that you did in `_validation_py.py`

ryan-williams:
great, thanks for the feedback. I added the test and the `_validation_py.py` version.

I also noticed that the `validate_union` case *doesn't* have this problem, though I previously thought it did; `True` is returned before the `if raise_errors` check.

let's see what the CIs think…

PR #208, opened 2018-05-28
add block_reader interface

ryan-williams:
[Apache Beam is interested in using fastavro instead of apache/avro](https://issues.apache.org/jira/browse/BEAM-2810), but needs an interface like I've added here, for seeking to and reading individual record-blocks in a file.

[I've published this as `fastavro-blocks:0.19.5.4` to test.pypi.org](https://test.pypi.org/project/fastavro-blocks/0.19.5.4/) (cf. [`blocks-test-pypi` branch](https://github.com/ryan-williams/fastavro/tree/blocks-test-pypi), which has [a commit on top of this PR changing the name and version](https://github.com/ryan-williams/fastavro/commit/e602bddefea6e3a369ddbd5c3695cf3d700da586)), and [used that in a Beam branch here](https://github.com/ryan-williams/beam/tree/fastavro).

Presumably it can't go into Beam until it's in real pypi, so that will have to wait until the next release here, though those seem to happen quite frequently 🎉😀

In this change I introduced a block-iteration interface, and had the standard record-iterator wrap it. I also did a few random tweaks to make _read{.pyx,_py.py} diff more cleanly against one another.

I'm new to Cython and haven't figured out when `_read_py.py` would be used instead of `_read.pyx`; a previous version of this PR failed Travis' pypy builds due to a bug I introduced in `_read_py.py`, so I guess it's covered?

### DIFF
--- a/fastavro/_read_py.py
+++ b/fastavro/_read_py.py
@@ -15,7 +15,7 @@ from uuid import UUID
 import json
 
 from .six import (
-    MemoryIO, xrange, btou, iteritems, is_str, str2ints, fstint
+    xrange, btou, iteritems, is_str, str2ints, fstint
 )
 from .schema import (
     extract_record_type, extract_named_schemas_into_repo, populate_schema_defs,
@@ -484,10 +484,23 @@ class ReaderBase(object):
         raise NotImplemented
 
 
+class MemoryReader(ReaderBase):
+    def __init__(self, data):
+        self.data = data
+        self.position = 0
+
+    def read(self, size):
+        result = self.data[self.position: self.position + size]
+        self.position += size
+        return result
+
+    def __len__(self):
+        return len(self.data)
+
+
 def null_read_block(fo):
     """Read block in "null" codec."""
-    read_long(fo)
-    return fo
+    return MemoryReader(read_bytes(fo))
 
 
 def deflate_read_block(fo):
@@ -495,7 +508,7 @@ def deflate_read_block(fo):
     data = read_bytes(fo)
     # -15 is the log of the window size; negative indicates "raw" (no
     # zlib headers) decompression.  See zlib.h.
-    return MemoryIO(decompress(data, -15))
+    return MemoryReader(decompress(data, -15))
 
 
 BLOCK_READERS = {
@@ -508,7 +521,7 @@ def snappy_read_block(fo):
     length = read_long(fo)
     data = fo.read(length - 4)
     fo.read(4)  # CRC
-    return MemoryIO(snappy.decompress(data))
+    return MemoryReader(snappy.decompress(data))
 
 
 try:
@@ -531,6 +544,14 @@ def acquaint_schema(schema):
 
 def _iter_avro(fo, header, codec, writer_schema, reader_schema):
     """Return iterator over avro records."""
+    for block in _iter_avro_blocks(fo, header, codec, writer_schema,
+                                   reader_schema):
+        for i in range(block.num_records):
+            yield read_data(block.bytes, writer_schema, reader_schema)
+
+
+def _iter_avro_blocks(fo, header, codec, writer_schema, reader_schema):
+    """Return iterator over avro blocks."""
     sync_marker = header['sync']
     # Value in schema is bytes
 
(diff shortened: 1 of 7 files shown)


scottbelden:
Thanks for the PR. I'll try to take a look soon and give a full review. One note is that we will want a test or two using the new functionality, but I can help with that if needed.

ryan-williams:
Thanks, I'm adding some tests and fixing bugs found along the way; I'll bump this when that's done

ryan-williams:
OK I think this is in a better state; I added a test, and cleaned up the `Block.{offset,size}` semantics so that they refer to actual file- or memory-positions and distances, whether the data is compressed or not.

ryan-williams:
Thinking I'll expand the tests to verify reading the records from a temp-file (currently they just read from an in-memory buffer)

fastavro/_read_py.py
```diff
@@ -478,34 +479,68 @@ def skip_sync(fo, sync_marker):
...
+    def read(self, size):
+        result = self.data[self._position: self._position + size]
+        self._position += size
+        return result
+
+    def __len__(self):
+        return len(self.data)
+
+
+class FileObjectReader(ReaderBase):
```
scottbelden: I know the cython side needed a special `Reader` object in the past because the underlying cython type couldn't be used in the same way as the python type. Is the need here on the python side basically just for being able to call `len()` (as it seems the `MemoryIO` object doesn't have a way to do that)?

fastavro/_read_py.py
```diff
@@ -478,34 +479,68 @@ def skip_sync(fo, sync_marker):
         raise ValueError('expected sync marker not found')
 
 
+class ReaderBase(object):
+    def __init__(self):
+        self._position = 0
+
+    def position(self):
```
scottbelden: Should this be changed to `tell` to make it more similar to a normal `io` object?

fastavro/_read_py.py
```diff
@@ -478,34 +479,68 @@ def skip_sync(fo, sync_marker):
...
 def null_read_block(fo):
     """Read block in "null" codec."""
-    read_long(fo, None)
-    return fo
+    return MemoryReader(read_bytes(fo))
 
 
 def deflate_read_block(fo):
     """Read block in "deflate" codec."""
-    data = read_bytes(fo, None)
```
scottbelden: Thanks! Been meaning to fix this

fastavro/_read_py.py
```diff
@@ -523,50 +558,63 @@ def acquaint_schema(schema):
...
-            for record in avro_reader:
-                process_record(record)
-    """
 
+class file_reader:
     def __init__(self, fo, reader_schema=None):
-        self.fo = fo
         try:
-            self._header = read_data(fo, HEADER_SCHEMA)
+            self._fo_reader = FileObjectReader(fo)
```
scottbelden: Can we move this to before the `try`? Just to make it clear that the `try` is only for the `read_data` call?

fastavro/_read_py.py
```diff
@@ -588,21 +636,84 @@ def __init__(self, fo, reader_schema=None):
...
+
+
+    Example::
+
+        from fastavro import block_reader
+        with open('some-file.avro', 'rb') as fo:
+            avro_reader = block_reader(fo)
+            schema = avro_reader.schema
+            for block in avro_reader:
+                process_block(record)
```
scottbelden: I think this should be `process_block(block)`, right?

fastavro/_read_py.py
```diff
@@ -32,7 +32,7 @@
 )
 
 MASK = 0xFF
-AVRO_TYPES = set([
+AVRO_TYPES = {
```
scottbelden: Thanks for the cleanup here

fastavro/_read_py.py
```diff
@@ -478,34 +479,68 @@ def skip_sync(fo, sync_marker):
...
+    return MemoryReader(decompress(data, -15))
 
 
 BLOCK_READERS = {
     'null': null_read_block,
     'deflate': deflate_read_block
 }
 
+
+def snappy_read_block(fo):
```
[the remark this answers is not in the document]
ryan-williams: I believe it was to minimize spurious diffs between `_read.pyx` and `_read_py.py`. 

as a sanity check I look at them in `meld` and eyeball the differences to make sure I didn't miss anything!

fastavro/_read_py.py
ryan-williams: yea, I didn't have a well-thought-through story here, thanks for calling out!

in the last two commits in the latest batch (https://github.com/tebeka/fastavro/pull/208/commits/2cfe013495aa41ea02feef50e3254c55067ceb18 and https://github.com/tebeka/fastavro/pull/208/commits/ca99dcb011803ed0627d03d86d9b08c9fb16dfdf), I took a stricter eye to which of the abstractions {`ReaderBase`, `MemoryReader`, `FileObjectReader`} were needed in each of {`.pyx`, `.py`}.

afaict, none of them are necessary!

The stream-like things we deal with are:
- "files" (which can be actual file-descriptors, or in-memory buffers simulating files in tests), and
- memory-buffers representing avro blocks.

The former need `read` and `tell` methods (`tell` is used for inferring block size after reading it), and the latter need only `read`. File-descriptors and `MemoryIO`s both support both.

I'm trying to run some of the benchmarks now to see whether there's a noticeable difference between the versions with or without these abstractions.

ryan-williams:
Thanks for all the feedback! I think I addressed everything, lmk what you think!

fastavro/_read_py.py
scottbelden: Great. That should reduce some complexity. We probably should have added a test if the original cython ones were needed, but it's possible it was done for some micro-optimizations. However, I think the reduced complexity and more similar to the python implementation are more beneficial.

scottbelden:
Thanks! This looks great. Merging this and will get a version up on pypi shortly.

swordspoet:
Nice job, but how can I directly get the length of each block?

ryan-williams:
@swordspoet IIRC, [`Block.bytes_`](https://github.com/fastavro/fastavro/blob/0.21.4/fastavro/_read_py.py#L566) is the uncompressed bytes, and [`Block.size`](https://github.com/fastavro/fastavro/blob/0.21.4/fastavro/_read_py.py#L572) is the compressed size, does that get you what you need?

(seems like they could use some docstrings…)

swordspoet:
@ryan-williams thank you, what I want is how many records in each block, and I want to improve speed by directly getting the numbers of record in each block. I have tried your method and got the size of block.

swordspoet:
@ryan-williams thank you, I found thishttps://github.com/fastavro/fastavro/blob/0.21.4/fastavro/_read_py.py#L567

scottbelden:
@swordspoet @ryan-williams I created an issue to better document those attributes (as well as others) when I find some time. Hopefully this will make it easier for others in the future.

PR #231, opened 2018-06-28
fix field name when validating arrays

scottbelden:
When running the following snippet...

```python
my_schema = {
    "fields": [
        {
            "name": "array",
            "type": {
                "type": "array",
                "items": "string",
            },
        },
    ],
    "namespace": "namespace",
    "name": "test_validate_array",
    "type": "record"
}

datum = {"array": [1]}
validate(datum, my_schema)
```

An exception with this message is produced:

```python
fastavro._validate_common.ValidationError: [
  " is <1> of type <class 'int'> expected string"
]
```

This change will make sure the field name shows up correctly like so:

```python
fastavro._validate_common.ValidationError: [
  "namespace.test_validate_array.array is <1> of type <class 'int'> expected string"
]
```

### DIFF
--- a/fastavro/_validation.pyx
+++ b/fastavro/_validation.pyx
@@ -81,13 +81,9 @@ cdef inline bint validate_array(datum, dict schema,
     if not isinstance(datum, Iterable) or is_str(datum):
         return False
 
-    if raise_errors:
-        namespace, name = schema_name(schema, parent_ns)
-    else:
-        name = parent_ns
     for d in datum:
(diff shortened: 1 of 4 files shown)


rhaarm:
Ah interesting catch, ya I've come across a few issues with non-named types causing issues. Hopefully this fixed all of those issues.

scottbelden:
Yeah, I want to add a few more tests and double check a few things before merging.

Issue #248, opened 2018-07-12
load_shcema error when Parent.avsc has multiple  Child type


lsyarn:
`test_reading_after_writing_with_load_schema` work fine when `Parent.avsc` has only on Child Type field(Defined by `Child.asvc`).But when `Parent.avsc` has multiple  Child type, an exception will be raised:`TypeError: 'dict' object cannot be interpreted as an integer`。

Use schemas bellow can reappear this bug:
Parent.avsc
```
{
  "type":"record",
  "name":"Parent",
  "fields":[
    {
      "name": "child",
      "type": "Child"
    },
    {
      "name": "anotherChild",
      "type": "AnotherChild"
    }
  ]
}
```
Child.avsc
```
{
  "type":"record",
  "name":"Child",
  "fields":[]
}
```
AnotherChild.avsc
```
{
  "type":"record",
  "name":"AnotherChild",
  "fields":[]
}
```

scottbelden:
Resolved via https://github.com/fastavro/fastavro/pull/249

PR #249, opened 2018-07-12
fix issue #248

lsyarn:
An `list.insert` call mistake, in '_schema._load_schema' function. Cause error when Parent schema has multiple user-define-type.


scottbelden:
Thanks! This looks great.

PR #281, opened 2018-10-08
when de-serialize union type write tuple of type name and data to be consistent with the serializer that support it

hershs:
I case of union type the serializer supports the following syntax to choose type name:
```
{
    'union_field`: (
        'com.mycompany.MyMessageType',
        {
            'field1': 'val1'
        }
    )
}
```
Change reader to support same format when de-serialize messages

### DIFF
--- a/fastavro/__init__.py
+++ b/fastavro/__init__.py
@@ -52,6 +52,7 @@ import fastavro.validation
 reader = iter_avro = fastavro.read.reader
 block_reader = fastavro.read.block_reader
 schemaless_reader = fastavro.read.schemaless_reader
+configure = fastavro.read.configure
 writer = fastavro.write.writer
 schemaless_writer = fastavro.write.schemaless_writer
 is_avro = fastavro.read.is_avro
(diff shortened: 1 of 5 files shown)


scottbelden:
I think this might need to be guarded by some configuration option as I think this would be a breaking change where users would now get back a tuple of the type and data rather than just the data. At the moment though I don't think we have a good system in place for configuration options so this will probably need some thought.

On the plus side, something like this would probably resolve https://github.com/fastavro/fastavro/issues/205

hershs:
@scottbelden , do you have already configuration in this project? How can I use it?

hershs:
@scottbelden , I'm new in python, do you have idea for configuration option?

scottbelden:
Not yet. I'll try to think about it some and see if I can find any other projects that do something similar to get inspiration from.

hershs:
In my scenario I read messages with several union records fields from kafka, do some enrichment in other fields and send it to another kafka topic. So, I send the union fields as fastavro reads them for me. When serialize it back it choose wrong record type from the union.
Therefore I need some solution for this scenario.

PR #293, opened 2018-10-30
Bugfix regression since 0.21.7: Allow writing to non-seekable streams again

jquast:
commit aaf0242308c4299bb2f8313e9e25485e8f657dfe  introduces a nice new feature to allow to writing to files that were opened in append mode.

However, this directly calls ``tell()`` without concern for whether the stream supports it. For network sockets (not tested), or, especially ``sys.stdout.buffer`` (tested manually), such streams do not have a position, and raise OSError.

We simply guard ourselves from an ``OSError`` by first calling the ``file_like.seekable()`` method in python 3, and expecting possible ``OSError, IOError`` python 2, and raising False in such case. If ``tell()`` raises exception, seek() most certainly will.

I have included simple unit tests that should provide full coverage, and here is example code using ``sys.stdout.buffer`` to write avro directly to stdout, showing how it worked previously (0.21.7), regressed (master), and works again with this PR.

```
import sys, fastavro
schema = {"type": "record", "name": "test_metadata", "fields": []}
fastavro.writer(sys.stdout.buffer, schema, [{}])
```

## Before fix (master) 0.21.10

```
~/Code/fastavro/fastavro/_six.pyx in fastavro._six.py3_appendable()
     48 
     49     def py3_appendable(file_like):
---> 50         if file_like.tell() != 0:
     51             if file_like.readable():
     52                 return True

OSError: [Errno 29] Illegal seek
```

## after fix (branch `suggest-seekable`)

success, writes to stdout:
```
Objavro.codenullavro.schemar{"type": "record", "name": "test_metadata", "fields": []}?`u?.???G?H/??`u?.??G?H/?
```

## from b8b4439

Note this is a regression, version 0.21.7, success, writes to stdout
```
Objavro.codenullavro.schemar{"type": "record", "name": "test_metadata", "fields": []}=�)y:P�=I��.k�=�)y:P�=I��.k�>>> 
```

Thank you for your consideration of this PR!

### DIFF
--- a/fastavro/six.py
+++ b/fastavro/six.py
@@ -47,7 +47,7 @@ if sys.version_info >= (3, 0):
         return datum[0]
 
     def py3_appendable(file_like):
-        if file_like.tell() != 0:
+        if file_like.seekable() and file_like.tell() != 0:
             if file_like.readable():
                 return True
             else:
--- a/fastavro/six.py
+++ b/fastavro/six.py
@@ -112,12 +112,12 @@ else:  # Python 2x
             file_like.mode
         except AttributeError:
             # This is probably some io stream so we rely on its tell() working
-            if (file_like.seekable()
-                    and file_like.tell() != 0
-                    and _readable(file_like)):
-                return True
-            else:
-                return False
+            try:
+                if file_like.tell() != 0 and _readable(file_like):
+                    return True
+            except (OSError, IOError):
+                pass
+            return False
 
         if "a" in file_like.mode:
             if "+" in file_like.mode:
--- a/fastavro/six.py
+++ b/fastavro/six.py
@@ -112,7 +112,9 @@ else:  # Python 2x
             file_like.mode
         except AttributeError:
             # This is probably some io stream so we rely on its tell() working
-            if file_like.tell() != 0 and _readable(file_like):
+            if (file_like.seekable()
+                    and file_like.tell() != 0
+                    and _readable(file_like)):
                 return True
             else:
                 return False
--- a/tests/test_fastavro.py
+++ b/tests/test_fastavro.py
@@ -3,7 +3,7 @@ from fastavro.read import _read as _reader
 from fastavro.write import _write as _writer, Writer
 from fastavro._schema_common import SCHEMA_DEFS
 
-from fastavro.six import MemoryIO, appendable
+from fastavro.six import MemoryIO
 
 import pytest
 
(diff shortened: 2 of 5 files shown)


scottbelden:
Thanks for submitting this and adding the tests (that were clearly lacking). I have just a few comments, but overall it looks good.

tests/test_fastavro.py
```diff
@@ -1723,3 +1723,53 @@ def test_appending_records_different_schema_fails(tmpdir):
...
+
+    with open(test_file, "a") as new_file:
+        new_file.write('this phrase forwards cursor position beyond zero')
+        with pytest.raises(ValueError) as exc:
+            appendable(new_file)
+        assert "you must use the 'a+' mode" in str(exc)
+
+
+def test_appendable_true_nonzero(tmpdir):
+    """six.appendable() returns false when file_like.tell() is non-zero."""
```
scottbelden: Shouldn't this say "returns True", not "returns false"?

tests/test_fastavro.py
```diff
@@ -1723,3 +1723,53 @@ def test_appending_records_different_schema_fails(tmpdir):
...
+            # mock a 0 position
+            return 0
+
+    assert not appendable(MockFileLike())
+
+
+def test_appendable_false_unseekable_stream():
+    """File streams that cannot seek return False."""
+
+    class MockFileLike:
```
scottbelden: Can you change this name to `MockStreamLike` as that seems to better describe it.

fastavro/six.py
```diff
@@ -112,10 +112,12 @@ def py2_appendable(file_like):
...
         except AttributeError:
             # This is probably some io stream so we rely on its tell() working
-            if file_like.tell() != 0 and _readable(file_like):
-                return True
-            else:
-                return False
+            try:
+                if file_like.tell() != 0 and _readable(file_like):
+                    return True
+            except (OSError, IOError):
```
scottbelden: In the tests you show the example of `sys.stdout.buffer` that raises `OSError`. Is there an example of something that will raise an `IOError`? Is this a python2 thing where it raised `IOError` there but on python3 it now raises `OSError`?

tests/test_fastavro.py
```diff
@@ -1723,3 +1723,53 @@ def test_appending_records_different_schema_fails(tmpdir):
...
+def test_appendable_true_nonzero(tmpdir):
+    """six.appendable() returns false when file_like.tell() is non-zero."""
+    test_file = str(tmpdir.join("test.avro"))
+
+    with open(test_file, "a+b") as new_file:
+        new_file.write(b'this phrase forwards cursor position beyond zero')
+        assert appendable(new_file)
+
+
+def test_appendable_false_zero(tmpdir):
```
scottbelden: I don't think we need the `tmpdir` fixture here, right?

tests/test_fastavro.py
```diff
@@ -1723,3 +1723,53 @@ def test_appending_records_different_schema_fails(tmpdir):
             fastavro.writer(new_file, different_schema, [{"field": 1}])
 
         assert "does not match file writer_schema" in str(exc)
+
+
+def test_appendable_raises_valuerror(tmpdir):
```
scottbelden: Do you mind moving these new tests into a new `test_six.py` file under the tests directory?

jquast:
I believe I addressed all of your concerns @scottbelden thanks for taking the time to review.

fastavro/six.py
scottbelden: Okay. Since this is the py2 section it should be safe to just catch `IOError` as the python3 case is handled in `py3_appendable`

scottbelden:
Thanks. Will be release in 0.21.11 shortly.

PR #306, opened 2018-12-07
Publish wheel for py27mu (in addition to py27m)

brianmartin:
We run `py27mu`, which requires installing via `.tar.gz` (and therefore
all the compilation dependencies are required as well). But we vastly
prefer the speed and simplicity of installing fastavro via wheel.

This change publishes a wheel for `py27mu`.

-----

The full list of distributions currently supported by the manylinux
image is:

```
$ docker run -it quay.io/pypa/manylinux1_x86_64 bash -c "ls /opt/python"
cp27-cp27m  cp27-cp27mu  cp34-cp34m  cp35-cp35m  cp36-cp36m  cp37-cp37m
```

All of these are currently published as wheels with the exception of
`cp27mu`.

I've tested this change by running this invocation from the `.travis.yml`:

```
 $ docker run --rm -v "$(pwd):/data" quay.io/pypa/manylinux1_x86_64 bash -c "/data/.travis_build_wheel.sh 2.7 && ls /data/dist/"
...
fastavro-0.21.14-cp27-cp27m-manylinux1_x86_64.whl
fastavro-0.21.14-cp27-cp27mu-manylinux1_x86_64.whl
```

------

Let me know if there's more I can help with!

### DIFF
--- a/.travis_build_wheel.sh
+++ b/.travis_build_wheel.sh
@@ -10,6 +10,11 @@ TRAVIS_PYTHON_VERSION=${TRAVIS_PYTHON_VERSION/.}
 /opt/python/*${TRAVIS_PYTHON_VERSION}*m/bin/pip install cython
 FASTAVRO_USE_CYTHON=1 /opt/python/*${TRAVIS_PYTHON_VERSION}*m/bin/python setup.py bdist_wheel
 
+# Also produce a wheel for wide-char distributions.
+if ls "/opt/python/*${TRAVIS_PYTHON_VERSION}*mu" &> /dev/null; then
+    FASTAVRO_USE_CYTHON=1 /opt/python/*${TRAVIS_PYTHON_VERSION}*mu/bin/python setup.py bdist_wheel
+fi
+
 # Fix wheel
 for whl in dist/*.whl; do
     auditwheel repair "$whl" -w dist
--- a/publish.sh
+++ b/publish.sh
@@ -19,15 +19,16 @@ OSes="win_amd64
 macosx_10_13_x86_64
 manylinux1_x86_64"
 
-PyVers="27
-34
-35
-36
-37"
+PyVers="27m
+27mu
+34m
+35m
+36m
+37m"
 
 for os in $OSes; do
     for pyver in $PyVers; do
-        wget -q --directory-prefix=dist/ https://github.com/fastavro/fastavro/releases/download/${ver}/fastavro-${ver}-cp${pyver}-cp${pyver}m-${os}.whl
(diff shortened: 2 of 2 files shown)


brianmartin:
Hey @scottbelden, for a bit more information on this "ABI", I've found that [this section of PEP513](https://www.python.org/dev/peps/pep-0513/#ucs-2-vs-ucs-4-builds) is the most straightforward:

> All versions of CPython 2.x, plus CPython 3.0-3.2 inclusive, can be built in two ABI-incompatible modes: builds using the --enable-unicode=ucs2 configure flag store Unicode data in UCS-2 (or really UTF-16) format, while builds using the --enable-unicode=ucs4 configure flag store Unicode data in UCS-4. (CPython 3.3 and greater use a different storage method that always supports UCS-4.) If we want to make sure ucs2 wheels don't get installed into ucs4 CPythons and vice-versa, then something must be done.

Given that it's so rare to support those early {py30, py31, py32}, effectively just py27 is applicable here.

And for some deeper history [PEP261](https://www.python.org/dev/peps/pep-0261/) lays out the need for "wide" unicode in the first place.

scottbelden:
> CPython 3.3 and greater use a different storage method that always supports UCS-4

I think that part basically answers my question. Thanks!

brianmartin:
Great. Thanks for accepting and for the quick turnaround!

scottbelden:
Hmmm... I noticed the CI wasn't pushing a `mu` build to the releases page and so I tried your command (`docker run --rm -v "$(pwd):/data" quay.io/pypa/manylinux1_x86_64 bash -c "/data/.travis_build_wheel.sh 2.7 && ls /data/dist/"`) and I'm only seeing this output: `fastavro-0.21.15-cp27-cp27m-manylinux1_x86_64.whl`

brianmartin:
Oh, interesting.. I'm seeing both:

```
 $ git show
commit 2a6a8b3bf307c58145340774e6705534df53412c (HEAD -> master, tag: 0.21.15, origin/master, origin/HEAD)
Merge: c3d2464 feb83c1
Author: Scott Belden <scottabelden+github.sb@gmail.com>
Date:   Mon Dec 10 14:24:13 2018 -0500

    Merge pull request #306 from brianmartin/wheel-for-wide-char

    Publish wheel for py27mu (in addition to py27m)

$ docker run --rm -v "$(pwd):/data" quay.io/pypa/manylinux1_x86_64 bash -c "/data/.travis_build_wheel.sh 2.7 && ls /data/dist/" | tail -n2
warning: no files found matching '*' under directory 'conda.recipe'
fastavro-0.21.14-cp27-cp27m-manylinux1_x86_64.whl
fastavro-0.21.14-cp27-cp27mu-manylinux1_x86_64.whl
```

scottbelden:
I think by removing the quotes in the `if` check it works (checking on CI now).

brianmartin:
The new release is working for me. Thanks again @scottbelden!

PR #324, opened 2019-03-29
Add Writer.write_block()

josephglanville:
This method is effectively a fast path for writing blocks with a matching schema to an OCF writer.
The primary use cases for this are very fast concatenation and recompressing existing OCF files, or doing both at once which is common in big data scenarios where you might want to combine many small files into one.

This could be additionally be exposed as a `block_writer` interface similar to `block_reader`

Note I have not added any safety checks as the system we are using this in already hashes the schemas of incoming OCF files before concatenating them but a similar approach could be added to the library (or warning labels added) - figured I would ask what you prefer.

Additionally this fixes up some small issues in separate commits, if you prefer I can move those into a different PR.

### DIFF
--- a/fastavro/_read.pyx
+++ b/fastavro/_read.pyx
@@ -689,7 +689,7 @@ class Block:
 
     def __str__(self):
         return ("Avro block: %d bytes, %d records, codec: %s, position %d+%d"
-                % (len(self.bytes), self.num_records, self.codec, self.offset,
+                % (len(self.bytes_), self.num_records, self.codec, self.offset,
                    self.size))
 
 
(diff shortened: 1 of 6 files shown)


scottbelden:
Thanks for the PR. I'll try to take a look soon.

scottbelden:
I finally got a chance to look at this. I think this is okay as is without a check on the schema. The `Writer` interface isn't really documented as the `writer` (note the difference in the capitalization of the `w`) is what I would expect most people would use and so I think it's okay to assume that if people want to use the more low level `Writer` then they have to take the extra precautions.

Thanks for the other small fixes as well.

PR #331, opened 2019-04-26
Add CLI tests, bugfix '-' (stream stdin), and new feature --metadata

jquast:
- Rename `tests/test_main_module.py` to `tests/test_main_cli.py`, and change
  the unit tests, which call main directly, to system integration tests,
  which call the 'fastavro' CLI through subprocess module, and provide
  more test coverage!
- Bugfix: python3 can stream stdin, like ``cat file.avro | fastavro -`` I think this previously worked for python 2 when this CLI was first authored.
- New feature, `--metadata' argument, to display avro metadata, (without the schema).
- Small nit: Display the exception *type* in CLI error output.


jquast:
Just one small py27 str/unicode fix remaining

scottbelden:
Thanks! I'll try to take a look at this and the issue you mentioned this weekend.

PR #340, opened 2019-05-12
add option to expand schemas

scottbelden:
Resolves https://github.com/fastavro/fastavro/issues/314


scottbelden:
Probably doesn't work yet...

PR #342, opened 2019-05-27
Fix Mac OSX Streaming output to sys.stdout.buffer (Closes #330).

### LINKED ISSUE #330: Mac/BSD, for sys.stdout: "When appending to an avro file you must use the 'a+' mode"
How curious how we learn how these platforms behave differently,

mac
---

```
$ python -V
Python 3.6.5
$ python -c 'import sys;print(sys.stdout.buffer.tell())'
1151335
```

linux
-----

```
$ python -V
Python 3.6.6
$ python -c 'import sys;print(sys.stdout.buffer.tell())'
OSError: [Errno 29] Illegal seek
```

This platform-specific behavior impedes an edge-case feature: what if we wanted to use fastavro to stream directly to `sys.stdout`? Such as in a classic unix pipe -- this works perfectly ok on Linux, great! 

Example:

```
In [3]: import sys, fastavro                                                                                         

In [4]: fout = fastavro.write.Writer(sys.stdout.buffer, schema=fastavro.parse_schema({'name': 'ohhai', 'type': 'record', 'fields': [ {'name': 'temp', 'type': 'int'}]})); fout.flush()                                            
Objavro.codenullavro.schema�{"type": "record", "name": "ohhai", "fields": [{"name": "temp", "type": "int"}], "__fastavro_parsed": true}D�A�7G+*+w�9ɨ�
```

Cool :) But on Mac, this isn't possible, the code introduced in https://github.com/fastavro/fastavro/commit/aaf02423#diff-fbbdb913f5928803608b2925d15d137fR49 raises the ValueError,

related comment, https://github.com/fastavro/fastavro/issues/276#issuecomment-424309799

```
In [2]: fout = fastavro.write.Writer(sys.stdout.buffer, schema=fastavro.parse_schema({'name': 'ohhai', 'type': 'record'
   ...: , 'fields': [ {'name': 'temp', 'type': 'int'}]})); fout.flush()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-dc0833ee58e4> in <module>
----> 1 fout = fastavro.write.Writer(sys.stdout.buffer, schema=fastavro.parse_schema({'name': 'ohhai', 'type': 'record', 'fields': [ {'name': 'temp', 'type': 'int'}]})); fout.flush()

~/Code/fastavro/fastavro/_write.pyx in fastavro._write.Writer.__init__()

~/Code/fastavro/fastavro/_six.pyx in fastavro._six.py3_appendable()

ValueError: When appending to an avro file you must use the 'a+' mode, not just 'a'
```

jquast:
- Create a TDD in ``tests/test_six.py``, where it is necessary to temporarily
  disable pytest's 'capfd' method, https://docs.pytest.org/en/latest/capture.html
- This test then fails, raising ValueError on OSX, and fixed by ``is sys.stdout`` identity check.
- To support this test, lazily bind to sys.stdout object reference in ``six.py`` (and ``_six.pyx``), we want sys.stdout reference at runtime rather than import, because we would otherwise get a handle of the "capfd" (file-like) stdout from pytest, even after disabling it our test.
- Unrelated but fun, use ``pytest.mark.parameterize()`` in ``tests/test_write_block.py``

Closes #330

### DIFF
--- a/tests/test_write_block.py
+++ b/tests/test_write_block.py
@@ -1,5 +1,9 @@
+# local
 import fastavro
 from fastavro.six import MemoryIO
+
+# 3rd party
+import pytest
 
 schema = {
     "type": "record",
@@ -46,7 +50,14 @@ def make_blocks(num_records=2000, codec='null'):
     return blocks, records
 
 
-def check_concatenate(source_codec='null', output_codec='null'):
+@pytest.mark.parametrize(
+    "source_codec,output_codec", [
+        ('null', 'null'),
+        ('deflate', 'deflate'),
+        ('null', 'deflate'),
+        ('deflate', 'null'),
+    ])
+def test_check_concatenate(source_codec, output_codec):
     blocks1, records1 = make_blocks(codec=source_codec)
     blocks2, records2 = make_blocks(codec=source_codec)
 
@@ -61,15 +72,3 @@ def check_concatenate(source_codec='null', output_codec='null'):
     new_file.seek(0)
     new_records = list(fastavro.reader(new_file, schema))
     assert new_records == records1 + records2
-
-
-def test_block_concatenation():
-    check_concatenate()
-
-
-def test_block_concatenation_deflated():
-    check_concatenate(source_codec='deflate', output_codec='deflate')
-
-
-def test_block_concatenation_deflated_output():
-    check_concatenate(source_codec='null', output_codec='deflate')
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+2019-05-27 version 0.21.24
+* Allow streaming to sys.stdout.buffer on OSX without raising ValueError (@jquast in PR #342)
+
 2019-05-06 version 0.21.23
 * Accept an ISO Date as a string (@scottbelden in PR #338)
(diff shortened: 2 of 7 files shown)


jquast:
If it's not too much to add, I'd also like to line-buffer stdout with the CLI in this same PR, so that when sent to a pipe it lets us plot things in realtime, like using "feedgnuplot" or something, thanks :-)

tests/test_write_block.py
```diff
@@ -46,7 +50,14 @@ def make_blocks(num_records=2000, codec='null'):
     return blocks, records
 
 
-def check_concatenate(source_codec='null', output_codec='null'):
+@pytest.mark.parametrize(
```
scottbelden: Thanks! This is definitely better than the way the test was previously being done.

tests/test_write_block.py
```diff
@@ -1,6 +1,10 @@
+# local
```
scottbelden: I'm probably going to remove this line and the `3rd party` as they are not present in all the other files (and in general I don't know if the code base always follows the convention of this import order)

PR #343, opened 2019-06-12
Another attempt at a JSON encoder/decoder

scottbelden:
Resolves https://github.com/fastavro/fastavro/issues/33
Resolves https://github.com/fastavro/fastavro/issues/69
Resolves https://github.com/fastavro/fastavro/issues/209


fastavro/_write_py.py
```diff
@@ -477,10 +495,14 @@ def __init__(self,
...
                  sync_marker=None):
-        self.fo = fo
-        self.schema = parse_schema(schema)
-        self.validate_fn = validate if validator is True else validator
-        self.io = MemoryIO()
+        GenericWriter.__init__(self, schema, metadata, validator)
+
+        self.metadata['avro.codec'] = codec
+        if isinstance(fo, BinaryEncoder):
+            self.encoder = fo
```
tvalentyn, line 503: After this change, the api for Writer class is different in pure Python and cythonized codepaths since
`Writer.fo` is no longer available in pure Python implementation, see https://github.com/apache/beam/pull/12927#discussion_r496712359. Is this intentional?

FYI, we (Apache Beam folks) are explicitly closing the associated file handle stored in Writer.fo after fastavro writes, as not doing so was causing issues  on some platforms: see https://issues.apache.org/jira/browse/BEAM-7747. Ideally, we'd expect the same API in both implementations.

tvalentyn: cc: @kamilwu, @chamikaramj

scottbelden: Yes, after adding in the JSON reader/writer the two diverged some in order to not break the other parts of the API that are more commonly used.

I will take a look and see if there's a clean way to make the `fo` attribute available in the same way on both implementations.

tvalentyn: Thank you!

PR #344, opened 2019-06-12
removed sys.stdout.buffer

DimaK415:
sys.stdout.buffer causes "AttributeError: 'OutStream' object has no attribute 'buffer'"


scottbelden:
What OS and python version are you using? Can you provide a minimal script to reproduce the error you are seeing?

scottbelden:
I'm going to close this as I'm not able to reproduce it and never heard back about the OS and python version that triggers it. If you have a way to reproduce it please let me know.

PR #349, opened 2019-06-19
Prepare logical types in validation

Artimi:
Attempt to fix the main issue in https://github.com/fastavro/fastavro/issues/173.

### DIFF
--- a/fastavro/_logical_writers.pyx
+++ b/fastavro/_logical_writers.pyx
@@ -1,0 +1,226 @@
+# cython: language_level=3str
+
+import datetime
+
+import decimal
+import uuid
+
+from libc.time cimport tm, mktime
+from cpython.int cimport PyInt_AS_LONG
+from cpython.tuple cimport PyTuple_GET_ITEM
+
+from fastavro import const
+from ._six import long, mk_bits
+from ._timezone import epoch
+
+ctypedef long long long64
+
+cdef long64 MCS_PER_SECOND = const.MCS_PER_SECOND
+cdef long64 MCS_PER_MINUTE = const.MCS_PER_MINUTE
+cdef long64 MCS_PER_HOUR = const.MCS_PER_HOUR
+
+cdef long64 MLS_PER_SECOND = const.MLS_PER_SECOND
+cdef long64 MLS_PER_MINUTE = const.MLS_PER_MINUTE
+cdef long64 MLS_PER_HOUR = const.MLS_PER_HOUR
+
+
+# The function datetime.timestamp() is a simpler, faster way to convert a
+# datetime to a Unix timestamp, but is only available in Python 3.3 and later.
+cdef has_timestamp_fn = hasattr(datetime.datetime, 'timestamp')
+
+
+cpdef long64 prepare_timestamp_millis(object data, schema):
+    cdef object tt
+    cdef tm time_tuple
+    if isinstance(data, datetime.datetime):
+        if not has_timestamp_fn:
+            if data.tzinfo is not None:
+                return <long64>(<double>(
+                    <object>(data - epoch).total_seconds()) * MLS_PER_SECOND
+                )
+            tt = data.timetuple()
+            time_tuple.tm_sec = PyInt_AS_LONG(<object>(PyTuple_GET_ITEM(tt, 5)))
+            time_tuple.tm_min = PyInt_AS_LONG(<object>(PyTuple_GET_ITEM(tt, 4)))
+            time_tuple.tm_hour = PyInt_AS_LONG(<object>(PyTuple_GET_ITEM(tt, 3)))
+            time_tuple.tm_mday = PyInt_AS_LONG(<object>(PyTuple_GET_ITEM(tt, 2)))
+            time_tuple.tm_mon = PyInt_AS_LONG(<object>(PyTuple_GET_ITEM(tt, 1))) - 1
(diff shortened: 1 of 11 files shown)


fastavro/write.py
```diff
@@ -11,7 +12,7 @@
 Writer = _write.Writer
 json_writer = json_write.json_writer
 schemaless_writer = _write.schemaless_writer
-LOGICAL_WRITERS = _write.LOGICAL_WRITERS
+LOGICAL_WRITERS = logical_writers.LOGICAL_WRITERS
```
Artimi: I imported it also here to keep the interface of `write` untouched.

tests/test_fastavro.py
```diff
@@ -1898,3 +1899,77 @@ def test_more_null_union_issues():
...
+        {"item": None},
+        {"item": {"item": "2019-05-06"}}
+    ]
+
+    expected = [
+        {"item": None},
+        {"item": {"item": datetime.date(2019, 5, 6)}}
+    ]
+
+    assert expected == roundtrip(schema, records)
```
Artimi: I'm not sure if I placed the tests correctly.

fastavro/_validation.pyx
```diff
@@ -79,7 +80,8 @@ cdef inline bint validate_float(datum, schema=None,
 cdef inline bint validate_fixed(datum, dict schema,
                                 str parent_ns='', bint raise_errors=True):
     return (
-        (isinstance(datum, bytes) and len(datum) == schema['size'])
+        ((isinstance(datum, bytes) or isinstance(datum, bytearray))
+         and len(datum) == schema['size'])
```
Artimi: `test_complex_schema` fails without this.

scottbelden:
@Artimi I got busy this weekend and didn't have a chance to take a look at this. I just did and I think the direction seems good. I think I have a few small comments, but before marking those but before mentioning those, I had a quick question about the intention of this change.

Based on the tests you added, it seems like the intention of this PR is to fix the issue when logical types are used in unions. However, the tests say `customLogicalType` which seems a bit inaccurate. However, I wrote a test that used a custom `IPAddress` type and it did work.

```python
def test_custom_logical_type():
    class IPAddress(object):
        def __init__(self, ipaddr):
            self.ip = ipaddr

    from fastavro.logical_writers import LOGICAL_WRITERS

    def prepare_ipaddress_write(data, schema):
        if isinstance(data, IPAddress):
            return data.ip
        else:
            return data

    LOGICAL_WRITERS["string-testipaddress"] = prepare_ipaddress_write

    schema = {
        "name": "test_date_as_string",
        "type": "string",
        "logicalType": "testipaddress",
    }

    data1 = IPAddress("1.1.1.1")
    binary = serialize(schema, data1)
    data2 = deserialize(schema, binary)
    # assert data2 == data1  # This assert would work if we had a custom logical reader
    assert data2 == data1.ip

    data1 = "1.1.1.1"
    binary = serialize(schema, data1)
    data2 = deserialize(schema, binary)
    # assert data2.ip == data1  # This assert would work if we had a custom logical reader
    assert data2 == data1
```

I just wanted to know if the intention of this PR was to resolve both of these issues (which it seems to do). If so, I think this is valuable and we should move forward on it.

Artimi:
@scottbelden Sorry about the confusion. The name `customLogicalType` is indeed inaccurate and should be something like `dateLogicalType` or something general.
Yes, the intention of the PR is to make logical types work also in unions. I think that your test would pass also on the current master branch because this is the way we already use our custom logical type. The only piece missing is the bug when a logical type is in a union.

scottbelden:
Thanks for the clarification. As mentioned, I will mark up a few comments, but overall I think it looks pretty good!

fastavro/_logical_writers.pyx
```diff
@@ -0,0 +1,234 @@
+# cython: language_level=3str
+
+"""Python code for writing AVRO files"""
```
scottbelden: I would remove this line and the comment lines below it. I don't think this line makes sense in this new file and I don't think the comment lines are accurate for this file either (as we had implemented logical types before the apache repo)

fastavro/_logical_writers_py.py
```diff
@@ -0,0 +1,198 @@
+# cython: auto_cpdef=True
+
+"""Python code for writing AVRO files"""
```
scottbelden: I would remove this line and the comment lines below it. I don't think this line makes sense in this new file and I don't think the comment lines are accurate for this file either (as we had implemented logical types before the apache repo)

fastavro/_logical_writers.pyx
```diff
@@ -0,0 +1,234 @@
...
+
+from libc.time cimport tm, mktime
+from cpython.int cimport PyInt_AS_LONG
+from cpython.tuple cimport PyTuple_GET_ITEM
+
+from fastavro import const
+from ._six import long, mk_bits
+from ._timezone import epoch
+
+CYTHON_MODULE = 1  # Tests check this to confirm whether using the Cython code.
```
scottbelden: This isn't needed. The test only checks this value in the `reader` and `writer`

fastavro/write.py
scottbelden: Makes sense 👍

tests/test_fastavro.py
scottbelden: I think this is an okay place for them. They could possibly go under `test_validation` but there's not a strong reason to move them there.

Artimi:
@scottbelden Thank you for the review. I made changes according to your suggestions. I also removed the confusing name of the logical type in tests.

PR #360, opened 2019-07-31
Added tests for decimal serialization

jancespivo:
Hi,
I added tests for decimal serialization / deserialization and it shows some flaws.
`Decimal("-0.0")` is encoded to `b'\x02\xfe'` which is strange. The test fails here:
`Decimal('-0.0') != Decimal('-0.002')`

I think it might be encoded to `b'\x02\x00'` (same as `Decimal('0.0')`)

In python:
```
>>> Decimal('0.0') == Decimal('-0.0')
True
```

I've prepared a fix for it: https://github.com/fastavro/fastavro/pull/361

### DIFF
--- a/tests/test_logical_types.py
+++ b/tests/test_logical_types.py
@@ -143,30 +143,42 @@ schema_bytes_decimal = {
     "namespace": "namespace",
     "type": "bytes",
     "logicalType": "decimal",
-    "precision": 15,
+    "precision": 20,
     "scale": 3,
 }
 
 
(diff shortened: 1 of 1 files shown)


scottbelden:
@jancespivo This shouldn't be necessary since https://github.com/fastavro/fastavro/pull/361 was merged, right? They are the same tests?

jancespivo:
@scottbelden Yes, it can be closed.

PR #370, opened 2019-10-07
Issue191 and others

bstockton:
When reading an input record like this:

 	{"id": 123, "payload": ("ApplicationCreated", {"applicationId": "123456789UT", "data": "..."})}

 the output looks like this:

 	{'id': 123, 'payload': {'applicationId': '123456789UT', 'data': '...'}}

 Which makes it impossible to tell whether the payload is a ApplicationCreated or ApplicationSubmitted record.
 We have added an option to fastavro.schemaless_reader to make it return the data in an unambiguous format.
 Specifically, anywhere in the schema where there is a union of records, the output is a pair with the
 record name as the first element, like ("ApplicationCreated", ...). So in this case the data should come back
 out exactly as it went in, like

  {"id": 123, "payload": ("ApplicationCreated", {"applicationId": "123456789UT", "data": "..."})}

 This new option should is disabled by default and we have tried to preserve backwards compatibility.


scottbelden:
Thanks for the PR. The idea is definitely something that I think would be useful. I'll try to find some time soon to take a look at the changes.

bstockton:
@scottbelden 
Thanks. Check out the test I added in test_schemaless.py for a quick example. It looks like the travis-ci build is failing because of line length. I have mine set to 120 but I can reformat it if needed. I also made some small formatting changes in _read.pyx that can be reverted as well.

scottbelden:
I apologize for letting this languish. I will try to find some time this weekend. Thanks for the patience.

scottbelden:
@bstockton I think this looks good. Could you revert all the formatting changes as and fix the line lengths so that travis will pass? Eventually I'll get around to using `black` and never have to worry about that anymore, but until then...

bstockton:
Will do that today or tomorrow. Thanks for taking the time to look this over. We use this at the company I work for so it will be great if this gets pulled in upstream so we don't have to maintain and build our own binaries.

bstockton:
OK. I finally got everything formatted correctly so Travis is passing. I ran over everything with autopep8 but had to make a bunch of manual changes.

scottbelden:
Merged. Should have a new release out shortly. Thanks again!

PR #394, opened 2020-02-20
pre-epoch timestamps on windows

jmgpeeters:
Works fine on Windows and Python 3.7. There is still an issue with 2.7, when it hits https://github.com/fastavro/fastavro/blob/master/fastavro/_logical_writers.pyx#L41 - which would need a similar fix. We don't support 2.7 anymore, though, so would prefer to fix separately to avoid risk.

Tested successfully on Linux and Python 2.7 and 3.7.
It does feel a bit awkward, so happy to discuss if you see any improvements.

### DIFF
--- a/fastavro/_logical_writers.pyx
+++ b/fastavro/_logical_writers.pyx
@@ -99,7 +99,7 @@ cpdef prepare_timestamp_micros(object data, schema):
                         <object>(data - epoch_naive).total_seconds()) * MCS_PER_SECOND
                     )
             else:
-                return <long64>(<double>(data.timestamp()) * MLS_PER_SECOND)
+                return <long64>(<double>(data.timestamp()) * MCS_PER_SECOND)
     else:
         return data
 
--- a/fastavro/_logical_writers.pyx
+++ b/fastavro/_logical_writers.pyx
@@ -3,6 +3,7 @@
 import datetime
 
 import decimal
+import os
 import uuid
 
 from libc.time cimport tm, mktime
@@ -11,7 +12,7 @@ from cpython.tuple cimport PyTuple_GET_ITEM
 
 from fastavro import const
 from ._six import long, mk_bits, int_to_be_signed_bytes
-from ._timezone import epoch
+from ._timezone import epoch, epoch_naive
 
 ctypedef long long long64
 
@@ -23,6 +24,7 @@ cdef long64 MLS_PER_SECOND = const.MLS_PER_SECOND
 cdef long64 MLS_PER_MINUTE = const.MLS_PER_MINUTE
 cdef long64 MLS_PER_HOUR = const.MLS_PER_HOUR
 
+cdef is_windows = os.name == 'nt'
 
 # The function datetime.timestamp() is a simpler, faster way to convert a
 # datetime to a Unix timestamp, but is only available in Python 3.3 and later.
@@ -33,6 +35,7 @@ cpdef prepare_timestamp_millis(object data, schema):
     cdef object tt
     cdef tm time_tuple
     if isinstance(data, datetime.datetime):
+        print(f'date: {data}')
         if not has_timestamp_fn:
             if data.tzinfo is not None:
                 return <long64>(<double>(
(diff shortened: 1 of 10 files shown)


scottbelden:
Thanks for this. I have a few comments and will add them in a moment.

Could you try removing the `.pyd` files and comment out these lines from the run-tests:

```
set FASTAVRO_USE_CYTHON=1
python setup.py build_ext --inplace
```

And then try running again. I would expect us to need some similar changes to `fastavro/_logical_writers_py.py` but I don't see any in the diff.

fastavro/_logical_writers.pyx
```diff
@@ -33,6 +35,7 @@ cpdef prepare_timestamp_millis(object data, schema):
     cdef object tt
     cdef tm time_tuple
     if isinstance(data, datetime.datetime):
+        print(f'date: {data}')
```
scottbelden: Please remove this

fastavro/_logical_writers.pyx
```diff
@@ -50,7 +53,17 @@ cpdef prepare_timestamp_millis(object data, schema):
             return mktime(& time_tuple) * MLS_PER_SECOND + <long64>(
                 int(data.microsecond) / 1000)
         else:
-            return <long64>(<double>(data.timestamp()) * MLS_PER_SECOND)
+            if is_windows:
+                if data.tzinfo is not None:
```
scottbelden: Could we add a small comment here saying why we need to special case this with a link to the bug tracker? Something like this:

`On Windows, timestamps before the epoch will raise an error. See https://bugs.python.org/issue36439`

tests/test_logical_types.py
```diff
@@ -137,6 +137,39 @@ def test_not_null_date():
     assert (data2['date'] == datetime.date(2017, 1, 1))
 
 
+schema_datetime = {
```
scottbelden: Can you move this schema inside of the test function?

tests/test_logical_types.py
```diff
@@ -137,6 +137,39 @@ def test_not_null_date():
...
+def test_ancient_datetime():
+    data1 = {
+        'timestamp-millis': datetime.datetime(1960, 1, 1),
+        'timestamp-micros': datetime.datetime(1960, 1, 1)
+    }
+    binary = serialize(schema_datetime, data1)
+    data2 = deserialize(schema_datetime, binary)
+
+    assert data1['timestamp-millis'] == data2['timestamp-millis'].replace(
+        tzinfo=None)
```
scottbelden: What does the `replace(tzinfo=None)` do?

fastavro/_logical_writers.pyx
jmgpeeters: Oh wow, sorry about this. I noticed + removed it when doing the Linux tests, but forgot to push.
I'll attend to the other remarks as well.

jmgpeeters:
> Thanks for this. I have a few comments and will add them in a moment.
> 
> Could you try removing the `.pyd` files and comment out these lines from the run-tests:
> 
> ```
> set FASTAVRO_USE_CYTHON=1
> python setup.py build_ext --inplace
> ```
> 
> And then try running again. I would expect us to need some similar changes to `fastavro/_logical_writers_py.py` but I don't see any in the diff.

Ah, yes, agreed - will do. I'll add a comment to the README.md as well about doing this. To make it more explicit, maybe there should eventually be two scripts, e.g. run-tests-python.sh and run-tests-cython.sh (+ *.cmd for Windows).

tests/test_logical_types.py
jmgpeeters: It turns a tz-aware datetime into a naive one, by simply remove the timezone. Quick example:

```
UTC = pytz.UTC
d_utc = datetime.datetime(2001, 1, 1, 0, 0, 0, 0, UTC)
d_naive = datetime.datetime(2001, 1, 1, 0, 0, 0, 0)

assert d_utc != d_naive
assert d_utc.replace(tzinfo=None) == d_naive
```

In this test, the data2 datetimes have the fastavro UTC tzinfo, so removing it allows for a simple comparison with data1.

scottbelden:
> Ah, yes, agreed - will do. I'll add a comment to the README.md as well about doing this. To make it more explicit, maybe there should eventually be two scripts, e.g. run-tests-python.sh and run-tests-cython.sh (+ *.cmd for Windows).

Thanks. What I should do is have the CI do this. Right now the pure python implementation is only checked on linux when it runs with pypy but I just need to update the CI do run the pure python for all platforms.

PR #397, opened 2020-02-26
add new public api to get the fullname of a schema

scottbelden:
Resolves https://github.com/fastavro/fastavro/issues/378


rnpridgeon:
👏

PR #404, opened 2020-03-08
fix: allow run pytests with arguments.

marcosschroh:
This `PR` allows to run specific tests, by default run all tests

```bash
# run all tests
./run-tests.sh

# run specific tests
./run-tests.sh tests/test_compression.py::test_optional_codecs
```


scottbelden:
I usually use the `-k` option to run a specific test but according to the pytest docs this seems like the more appropriate way to run the specific test. Thanks!

PR #407, opened 2020-03-19
fix bug with union in reader and record in writer (fastavro#406)

eino:
Fix case when reader has an union (containing a record),
and writer has directly the record:
- try to continue matching union elements if the first one didn't match
- read with the matching reader record

### DIFF
--- a/fastavro/_read_py.py
+++ b/fastavro/_read_py.py
@@ -87,8 +87,12 @@ def match_schemas(w_schema, r_schema):
         # If the reader is a union, ensure one of the new schemas is the same
         # as the writer
         for schema in r_schema:
-            if match_types(w_schema, schema):
-                return True
+            try:
+                if match_types(w_schema, schema):
+                    return True
+            except SchemaResolutionError:
+                # Try to match the next schemas in the union
+                pass
         else:
             raise SchemaResolutionError(error_msg)
     else:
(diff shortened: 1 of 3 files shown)


fastavro/_read_py.py
```diff
@@ -326,6 +330,15 @@ def read_record(decoder, writer_schema, reader_schema=None,
...
     else:
+        # Handle case where reader_schema is an union
+        if isinstance(reader_schema, list):
+            for schema in reader_schema:
+                try:
+                    if match_types(writer_schema, schema):
+                        reader_schema = schema
+                except SchemaResolutionError:
+                    # Try to match the next schemas in the union
+                    pass
```
ryanhaarmann: Add the raise `for/else` statement like in `match_schemas` function

eino: you're right, I added the condition

ryanhaarmann: So the `else` will be evaluated every time, you need a `break` after finding the first `match_types()`.

eino: indeed. fixed, thanks !

scottbelden:
@eino Thanks for starting this. I think this looks good but I'm worried that we are duplicating some effort (specifically looping through the list again to check for the right schema).

I have a branch that I played with that instead would have `match_schemas` return the appropriate reader schema (instead of just a `True` value which we currently don't use anyways). Here is the branch: https://github.com/fastavro/fastavro/compare/issue406?expand=1

Let me know what you think. Thanks again!

eino:
@scottbelden thanks for your reply, your branch is indeed more efficient, and works as well.

If you want, I already wrote a unit test for this case in my branch, let's merge it with this fix? I opened another pull request (https://github.com/fastavro/fastavro/pull/410), and we can close this one.

Thanks !

scottbelden:
Resolved in https://github.com/fastavro/fastavro/pull/411

I'll try to get a release out shortly. Thanks again!

Issue #418, opened 2020-04-17
Error writing JSON records with schemas that include dependent schemas


gudjonragnar:
I have an Avro spec that I want to use when writing records as JSON. This spec contains schemas that have dependencies. When trying to write records using `json_writer` I however faced this:
```
Traceback (most recent call last):
  File "/Users/gudjonragnar/fastavro/tests/test_json.py", line 493, in test_with_dependent_schema
    new_records = roundtrip(schema, records)
  File "/Users/gudjonragnar/fastavro/tests/test_json.py", line 13, in roundtrip
    json_writer(new_file, schema, records)
  File "/Users/gudjonragnar/fastavro/fastavro/json_write.py", line 46, in json_writer
    return writer(AvroJSONEncoder(fo), schema, records)
  File "/Users/gudjonragnar/fastavro/fastavro/_write_py.py", line 595, in writer
    codec_compression_level,
  File "/Users/gudjonragnar/fastavro/fastavro/_write_py.py", line 476, in __init__
    self.encoder.configure(self.schema)
  File "/Users/gudjonragnar/fastavro/fastavro/io/json_encoder.py", line 69, in configure
    self._parser = Parser(schema, self.do_action)
  File "/Users/gudjonragnar/fastavro/fastavro/io/parser.py", line 14, in __init__
    self.stack = self.parse()
  File "/Users/gudjonragnar/fastavro/fastavro/io/parser.py", line 17, in parse
    symbol = self._parse(self.schema)
  File "/Users/gudjonragnar/fastavro/fastavro/io/parser.py", line 31, in _parse
    production.insert(0, self._parse(field["type"]))
  File "/Users/gudjonragnar/fastavro/fastavro/io/parser.py", line 95, in _parse
    raise Exception("Unhandled type: {}".format(record_type))
Exception: Unhandled type: test.Dependency
```
where `test.Dependency` is a schema that is already loaded and is featured in a field in another schema as so:
```
...
    "fields": [{
        "name": "_dependency",
        "type": "Dependency"
    }]
...
```
I wrote a simple test, using the `roundtrip()` method from `test_json.py` which fails with the above error.
```
from fastavro.schema import parse_schema

 def test_with_dependent_schema():
    """Tests a schema with dependent schema"""
    dependency = {
        "type": "record",
        "name": "Dependency",
        "namespace": "test",
        "fields": [{
            "name": "_name",
            "type": "string"
        }]
    }

    schema = {
        "type": "record",
        "name": "Test",
[...]

PR #419, opened 2020-04-17
fix error writing json with dependent schemas (#418)

gudjonragnar:
- Add a check for known schema types to `parser.py`.
- Add a test to `test_json.py` that tests writing a record to json with a schema that includes a dependency on another known schema.

### DIFF
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -1,4 +1,5 @@
 from fastavro import json_writer, json_reader
+from fastavro.schema import parse_schema
 from fastavro.six import StringIO
 
 import json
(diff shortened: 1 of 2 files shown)


scottbelden:
Thanks! I can't believe I never had a test that included a simple named type. I just have one small comment about the test but otherwise this looks great.

tests/test_json.py
```diff
@@ -451,3 +452,44 @@ def test_union_in_map():
...
+    }
+
+    records = [{
+        '_name': 'parent',
+        '_dependency': {
+            '_name': 'child'
+        }
+    }]
+
+    parse_schema(dependency, "/tmp/dir")
```
scottbelden, line 491: I don't think the second argument here is doing what you expect. The second argument should be `True`/`False` and is an option to `expand` the schema. It's only useful when you want to save off the resulting schema. Also, [as per the docs](https://fastavro.readthedocs.io/en/latest/schema.html#fastavro._schema_py.parse_schema), the option when used should be used as a keyword argument because that will be enforced in the future.

I would just change this and the line below to remove the second argument:

```suggestion
    parse_schema(dependency)
```

gudjonragnar: Thanks for the correction, I think it might be a leftover because I started out with `_load_schema` or some other function that I can't recall at the moment.

tests/test_json.py
```diff
@@ -451,3 +452,44 @@ def test_union_in_map():
...
+
+    records = [{
+        '_name': 'parent',
+        '_dependency': {
+            '_name': 'child'
+        }
+    }]
+
+    parse_schema(dependency)
+    parse_schema(schema, "/tmp/dir")
```
gudjonragnar, line 492:
```suggestion
    parse_schema(schema)
```

scottbelden:
Thanks. I'll try to get a new release out soon.

PR #428, opened 2020-04-29
Minor testing tweaks

dacjames:
I've been working on adding several features to fastavro (more PRs to come) and encountered a couple minor inconveniences.

The teardown was failing with AttributeError on captured.err. Not sure why, but treating it as a tuple fixed the problem.

The pandas change is just a quality of life improvement, skipping instead of failing the test for optional functionality.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,8 +12,8 @@ def look_for_cython_error(capfd):
     will usually print a message to stderr saying that an exception was
     ignored. Here we check for that in every test function."""
     yield
-    captured = capfd.readouterr()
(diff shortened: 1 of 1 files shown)


scottbelden:
Interesting. I don't think we were ever using that `capfd` correctly as docs say (and have said for many versions) that`readouterr` returns an `(out, err)` tuple. Thanks for catching that.

However, I don't think we should do the check for `pandas`. It should get installed on the CI, but if for some reason it doesn't get installed I would want an error rather than it skipping the test.

dacjames:
No problem. I know there are different schools of thought on skipping tests. I removed it from the PR.

PR #429, opened 2020-04-29
Add missing fastavro/_logical_writers.c to Makefile

dacjames:
This looks to be a trivial oversight from a refactoring, as far as I can tell.


scottbelden:
Thanks!

PR #430, opened 2020-04-29
Special handling for unions containing both float and double

dacjames:
We have encountered a problem with union types that contain both 'float' and 'double'. The problem occurs writing a python float into a union type `['float', 'double']`. fastavro processes each type in the union in order, finds that the float matches, and narrows the python float (64 bits) to an avro float (32 bits), resulting in a loss of precision. Reading is not an issue because widening a float to a double is lossless.

This problem only occurs when writing and only in Python because of the ambiguity arising from having a single float type. We considered several possible solutions to this problem:

1. Never narrow python floats to avro floats. This solves the problem because the 'float' type in the union will no longer match, but it leaves python with no way to write to "float" type. In general, implicit narrowing is not great, but the lack of a way to explicitely narrow floats when needed ruled out this option.
2. Just fix the schema. One could argue that such union types are simply wrong and the library should not handle this case. Unfortunately, we have several of these schemas in the wild that are not easy to change. Moreover, our system also has C and Java producers and these schemas are handled there since these languages have explicit float/double types so there is no ambiguity.
3. Handle the issue in the `write_union` function. This works but was rejected for performance reasons. We parse a schem once and use it to write many times, so handling the issue when parsing the schema is beneficial.

I do not love modifying the schema in flight, so I'm open to any suggestions of an alternate solution. It's understandable for the fastavro project to take option 2 and not solve the problem at all but this solution works so I figured I might as well share just in case.

### DIFF
--- a/fastavro/_schema.pyx
+++ b/fastavro/_schema.pyx
@@ -67,14 +67,37 @@ def parse_schema(schema, expand=False, _write_hint=True, _force=False):
         return _parse_schema(schema, "", expand, _write_hint, set())
 
 
+cdef _reorder_float_unions(schema):
+    """Re-orders union types containing both "float" and "double",
+    so that "double" is preferred and no precision is lost when writing
+    64 bit python floats to ["float", "double"] union types
+    """
+
+    left = []
+    right = []
+
+    for sub in schema:
+        if sub == "float" or (hasattr(sub, "get") and sub.get("type") == "float"):
+            right.append(sub)
+        elif sub == "double" or (hasattr(sub, "get") and sub.get("type") == "double"):
+            left.append(sub)
+        else:
+            if len(right) == 0:
+                left.append(sub)
+            else:
+                right.append(sub)
+
+    return left + right
+
+
 cdef _parse_schema(schema, namespace, expand, _write_hint, named_schemas):
     # union schemas
     if isinstance(schema, list):
-        return [
+        return _reorder_float_unions([
             _parse_schema(
                 s, namespace, expand, False, named_schemas
             ) for s in schema
-        ]
+        ])
 
     # string schemas; this could be either a named schema or a primitive type
     elif not isinstance(schema, dict):
--- a/fastavro/_schema_py.py
+++ b/fastavro/_schema_py.py
@@ -194,14 +194,39 @@ def parse_schema(schema, expand=False, _write_hint=True, _force=False):
         return _parse_schema(schema, "", expand, _write_hint, set())
 
 
+def _reorder_float_unions(schema):
+    """Re-orders union types containing both "float" and "double",
+    so that "double" is preferred and no precision is lost when writing
+    64 bit python floats to ["float", "double"] union types.
+    """
+
+    left = []
+    right = []
+
+    for sub in schema:
+        if (sub == "float" or
+           (hasattr(sub, "get") and sub.get("type") == "float")):
+            right.append(sub)
+        elif (sub == "double" or
+              (hasattr(sub, "get") and sub.get("type") == "double")):
+            left.append(sub)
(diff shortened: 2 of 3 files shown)


scottbelden:
Sorry that it has taken me a while to provide any response to this. I think the implicit re-ordering could catch people off guard. While perhaps just as disruptive as changing the schema, I think [using the tuple-notation when writing](https://github.com/fastavro/fastavro/pull/153) can solve this too.

dacjames:
I can understand the objection to the implicit reordering but are you interested in fixing this problem at all? I think it's reasonable to expect that writing to a python float with type `['float', 'double']` would result in an avro `'double'`. As is, this footgun that can easily go unnoticed because it is visible only as a loss of precision.

Using tuple notation when writing would be too much breakage for us and doesn't remove the footgun for you.

scottbelden:
> I think it's reasonable to expect that writing to a python float with type ['float', 'double'] would result in an avro 'double'

I'm not sure I agree with this... The schema says it can be either a float or a double. To me, it feels intuitive that the first matching schema should be selected. If the schema is defined as `['float', 'double']` then it would match `float` and if it was defined as `['double', 'float']` then it would match `double`.

While I have not verified it yet, I believe this same scenario could come up with a union of a `string` and an `enum`. If the enum has symbol `FOO` and the user submits the string `"FOO"` then it's not really possible to know if that should match `string` or the `enum`. I would argue that the same rule of the first matching schema should apply.

If this were to be merged, not only would it be a breaking change (though I am not sure how often a `union['float', 'double']` is used), but it make it so that if someone actually did want to choose `float` then their only option would be to use the tuple notation when writing (whereas right now their options are using the tuple notation or changing the order of the union).

I understand that in your case changing the schema order or using the tuple notation are not viable options, but I don't think always choosing `double` is what everyone is expecting.

dacjames:
The situation with enum is not analogous because encoding "FOO" as an enum is lossless; I get all the information I wrote back when reading. I agree that processing the schema in order make sense in general.

To me, matching avro `'float'` for Python floats is the root of the problem. Python has no 32bit floating point type, so doing so is technically not correct. Outside of unions, there is no viable alternative, leaving you no option for supporting avro `'float'` but to implicitly narrow the value. Within a union, there is a viable alternative, which is to not match`'float'` if a `'double'` is available. You might say that 'float' is a "lossy" match, which you should only accept if no "lossless" match is available.

When I'm writing a floating point number with 64 bits of precision (as Python floats do) to the schema `['float', 'double']`, I would expect the `'double'` type to match, not the `'float'`. That's how it works in other languages: if I'm writing a C or Java `double` the avro `'double'` type is selected. As far as I'm aware, only floats have this silent narrowing behavior: if I try to write an oversized integer, fastavro fails instead.

I can tell you from experience that the current behavior is not intuitive. IIRC, it went months without being noticed, resulting in permanent data loss. Keep in mind that it's common in a larger organization for people to use schemas they did not write to conform to requirements for a downstream system so they're not looking closely at the ordering of types in a union. Python developers, in particular, are accustomed to having arbitrary precision which makes information loss like this is even more surprising.

I can't argue the compatibility point. I don't see any practical use for the current behavior but [Hyrum's Law](https://github.com/dwmkerr/hacker-laws#hyrums-law-the-law-of-implicit-interfaces) still holds.

scottbelden:
That's a fair point about lossy aspect of `['float', 'double']` versus the other cases like enum/string or bytes/fixed which lossless. However, I still don't feel comfortable with forcing the choice of `double`. I will make an issue about this so that if others stumble on this then they can weigh in. Unfortunately, any feedback will probably be one sided as those who are okay with the current behavior probably won't see or comment on the issue.

PR #431, opened 2020-04-29
Add a new logical type for fixed size signed and unsigned integers

dacjames:
Adds two new logical types: `sized-int` and `sized-uint`. These types allow for a schema like 
`{"type": "fixed", "size": 1, "logicalType": "sized_uint"}` which is equivalent to the C level `uint8_t` type. Notice that the logical type only indicates signed versus unsigned, since the size is already handled by the underlying fixed type.

This PR contains both a pure python implementation using the `struct` package and a hand-written cython implementation. The latter implementation is what we rely on in practice so care was taken to make it as efficient as practical. The special casing this required for Python versions is unfortunate and I would love to find an equally efficient solution that did not require it.

You may recall a similar PR a long time ago for the same feature, but based on the "name" rather than "logicalType". Our goal is the same, but the logical types feature made this addition much cleaner.

Happy to squash commits if that is preferred.

### DIFF
--- a/fastavro/_logical_writers.pyx
+++ b/fastavro/_logical_writers.pyx
@@ -275,7 +275,7 @@ cdef prepare_fixed_sized_uint(data, schema):
 
     if size == 1:
         d1 = <unsigned char> data
-        output[i] = <char> d1
+        output[0] = <char> d1
     elif size == 2:
         d2 = <unsigned short> data
         for i in range(0, size):
--- a/fastavro/_logical_writers.pyx
+++ b/fastavro/_logical_writers.pyx
@@ -9,6 +9,7 @@ import uuid
 from libc.time cimport tm, mktime
 from cpython.int cimport PyInt_AS_LONG
 from cpython.tuple cimport PyTuple_GET_ITEM
+from cpython.array cimport array, clone
 
 from fastavro import const
 from ._six import long, mk_bits, int_to_be_signed_bytes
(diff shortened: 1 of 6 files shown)


scottbelden:
I appreciate the work here, but I'm not really sure about this. This isn't a logical type defined in the specification and therefore wouldn't be supported in any other libraries. For the moment this feels to me like it should continue to exist in a fork of this library or be proposed as a new logical type to be added to the avro specification.

Theoretically in the past there were [some changes that should have helped to support custom logical types](https://github.com/fastavro/fastavro/issues/173) but I'd have to take a closer look to remember exactly how it worked (as unfortunately that PR didn't include a test using a custom type). It also possible this might not be sufficient as I'm assuming that PR doesn't allow for the custom logical types to be handled by custom cython code.

scottbelden:
Sorry, I don't think I explained my thoughts very well. You are right that other libraries should ignore any custom logical type. My fear is that this is a very specific use case that won't be used by anyone else. That could easily open the door adding and supporting any number of custom logical types.

dacjames:
I understand what you're saying but I guess I don't view the growth in custom logical types as a bad thing, even if they are rarely used.

Is there anything I could do to offset your concerns? I tried to make the change as isolated as possible but would some additional separation into something like a "contrib" package change your thinking?

scottbelden:
From the Java implementation it seems like they have a way of registering custom logical types: https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java#L38-L51

Something like that would be my preference.

dacjames:
Fair enough. For us, that's not worth the effort over a fork because of the complexity of adding cython downstream.

PR #438, opened 2020-05-20
Add type stubs

### LINKED ISSUE #434: Use type annotations
Type annotations are getting increasingly popular and common in the Python ecosystem. I'm interested in adding annotations to fastavro. Would a PR that does so be of interest?

spenczar:
This is an attempt at adding type stub information to fastavro. I've only attempted to add type information for the public APIs.

Since type information is in `.pyi` files, this doesn't affect Python 2 compatibility. On the other hand, there is a real risk that the type information is incorrect; `mypy` will simply blindly trust that the actual python code provided by modules will fulfill these interfaces.

The `block_reader` and `reader` classes in the read module are of particular note. In reality, those classes are subclasses of `file_reader`, but the type stubs don't indicate this. That's because `file_reader`'s `__iter__`, `next`, and `__next__` methods are complexly typed: they spit out elements of `self._elems`, which never gets actually set in a `file_reader`, and which has different types depending on the subclass. For a `block_reader`, they're `Block`s, and for a `reader`, they are `Message`s. I decided that conveying this is too complicated for now.

In addition, I made a helper convenience type definition in `types.pyi`. `AvroMessage` attempts to cover all the types that can be serialized or deserialized - not just `dicts` for Records, but I think plain values can be serialized too. That thing is worth reading over to make sure it's correct.

Fixes #434.


spenczar:
@scottbelden, I'm sorry, but I don't understand codecov's complaint here. It appears to believe that tests didn't run to cover missing codec libraries, but I didn't change that.

scottbelden:
Thanks. I'll try to take a look at this soon. And I wouldn't worry too much about the code coverage thing. Sometimes it gets all wacky and I'm not exactly sure how to make it more consistent.

Issue #424, opened 2020-04-26
fastavro doesn't parse protocols


debradley:
Currently, fastavro can parse individual schemas (avsc) but not collections of schemas collected inside a protocol (avpr). Because my projects use protocols, I have to continue to pull in avro-python3 as a dependency solely for parsing protocols even when I use fastavro for all my other avro needs. I would like to be able to drop the avro-python3 dependency altogether, so I'm proposing adding protocol parsing to fastavro as a new feature.

Are there philosophical or technical reasons that fastavro wouldn't support this? I may be able to work towards a PR if there's no opposition and no effort already underway.

scottbelden:
I haven't looked at protocols very closely. However, I don't believe there's any reason it was specifically not supported. I think it's just a result of there not being any requests for the feature.

PR #441, opened 2020-06-23
Convert iterables to list in cython array writing

ksunden:
In particular, I was trying to write tuples as arrays (the closest avro type).

Writing tuples (or any iterable) as an array already works in the pure python implementation, this makes the cython behavior in line with the python behavior.

P.S. I am implementing [Avro-RPC](https://gitlab.com/yaq/yaqd-core-python/-/tree/avro/yaqd_core/avrorpc) (and protocol reading #424 on top of fastavro). Currently still in early stages, not 100% standard compliant. I also use some 3.7+ features in my current implementation, but would be willing to clean it up/make it work on fastavro supported pythons once I have the finish the implementation.

### DIFF
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -326,6 +326,8 @@ cpdef write_data(bytearray fo, datum, schema):
     elif record_type == 'enum':
         return write_enum(fo, datum, schema)
     elif record_type == 'array':
+        if not isinstance(datum, list):
+            datum = list(datum)
         return write_array(fo, datum, schema)
     elif record_type == 'map':
         return write_map(fo, datum, schema)
--- a/tests/test_complex.py
+++ b/tests/test_complex.py
@@ -122,3 +122,8 @@ def test_complex_schema_nulls():
         {'multi_union_time': None, 'array_bytes_decimal': None,
          'array_fixed_decimal': None, 'union_uuid': None})
     assert (data1_compare == data2)
+
+def test_array_from_tuple():
+    data_list = serialize({"type": "array", "items": "int"}, [1,2,3])
+    data_tuple = serialize({"type": "array", "items": "int"}, (1,2,3))
+    assert data_list == data_tuple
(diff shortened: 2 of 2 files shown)


scottbelden:
Can you create or show a test that will fail without this? I'm having trouble understanding the problem.

ksunden:
I can add this as a test in a bit, but here is a quick example:

```python
>>> import fastavro
>>> import io
>>> fo = io.BytesIO()
>>> fastavro.schemaless_writer(fo, {"type": "array", "items": "int"}, [1,2,3]) # works
>>> fastavro.schemaless_writer(fo, {"type": "array", "items": "int"}, (1,2,3)) # fails
  File "<stdin>", line 1, in <module>
  File "fastavro/_write.pyx", line 635, in fastavro._write.schemaless_writer
    write_data(tmp, record, schema)
  File "fastavro/_write.pyx", line 329, in fastavro._write.write_data
    return write_array(fo, datum, schema)
TypeError: Expected list, got tuple
>>> fastavro._write_py.schemaless_writer(fo, {"type": "array", "items": "int"}, (1,2,3)) # works
```

scottbelden:
Okay. This seems fine but I will wait until the test is in place to merge. However, it should be noted that when reading records, `array` fields will always come back as a list, not a tuple.

ksunden:
Yes of course, the encoded values/schema are identical for a python list or a tuple (or other iterator), there is no way to determine that it started as anything other than a list.

The distinction goes away over the serialization/deserialiation process, but good for interfaces to be flexible in their allowed types, in general (when there is an obvious "correct" handling, as here when the schema is known)

ksunden:
I will note that tuples have a meaningful to fastavro itself distinction for union types, and that behavior is unchanged here, this only apply when encoding to arrays which are not part of a union (well, it applies to the data part of a tuple there, e.g. `('name', (1, 2, 3))` will work)

scottbelden:
Thanks!

PR #451, opened 2020-07-16
do not use global SCHEMA_DEFS

scottbelden:
Resolves https://github.com/fastavro/fastavro/issues/450

This is a work in progress... We'll see how it goes.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,8 +44,7 @@ jobs:
     displayName: 'Install dependencies'
 
   - script: |
-      ./run-tests.sh
-      codecov
+      ./run-tests.sh && codecov
     displayName: 'pytest'
 
   - script: |
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -71,7 +71,7 @@ def read_schemaless(iostream, schema, num_records, runs=1):
             end = time.time()
             times.append(end - start)
     print('... {0} runs averaged {1} seconds'.format(runs, (sum(times) / runs)))
-    return records
+    return record
 
 
 small_schema = {
--- a/fastavro/_read.pyx
+++ b/fastavro/_read.pyx
@@ -307,7 +307,6 @@ cdef read_fixed(
 ):
     """Fixed instances are encoded using the number of bytes declared in the
     schema."""
-    named_schemas[writer_schema["name"]] = writer_schema
     return fo.read(writer_schema['size'])
 
 
@@ -315,7 +314,6 @@ cdef read_enum(fo, writer_schema, named_schemas, reader_schema=None):
     """An enum is encoded by a int, representing the zero-based position of the
     symbol in the schema.
     """
-    named_schemas[writer_schema["name"]] = writer_schema
     index = read_long(fo)
     symbol = writer_schema['symbols'][index]
     if reader_schema and symbol not in reader_schema['symbols']:
@@ -454,11 +452,6 @@ cdef read_union(
     index = read_long(fo)
     idx_schema = writer_schema[index]
 
-    # Load named schemas
-    for schema in writer_schema:
-        if extract_record_type(schema) in ('record', 'enum', 'fixed'):
-            named_schemas[schema["name"]] = schema
-
     if reader_schema:
         # Handle case where the reader schema is just a single type (not union)
         if not isinstance(reader_schema, list):
@@ -534,7 +527,6 @@ cdef read_record(
          writer's schema does not have a field with the same name, then the
          field's value is unset.
     """
-    named_schemas[writer_schema["name"]] = writer_schema
     record = {}
     if reader_schema is None:
         for field in writer_schema['fields']:
@@ -922,6 +914,7 @@ class file_reader:
             self.reader_schema = parse_schema(
                 reader_schema,
                 _write_hint=False,
+                _named_schemas=self._named_schemas,
             )
         else:
             self.reader_schema = None
@@ -930,6 +923,7 @@ class file_reader:
             self._schema,
             _write_hint=False,
             _force=True,
+            _named_schemas=self._named_schemas,
         )
 
         self._elems = None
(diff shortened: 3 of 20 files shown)


scottbelden:
So far this seems to work. One aspect I need to look into is if it is even necessary to pass a dictionary in when calling `parse_schema` to get the named schemas. It seems like it shouldn't be necessary (at least when writing... maybe it is when reading).

Also, it would eventually be nice to stop using a bunch of individual read_type/write_type functions and have them all in a class. If that were the case, there would be no need to pass things like `named_schemas` around to all of those functions.

ksunden:
Is there an answer for named types being defined outside of the individual schema being written/read?

Currently with the global `SCHEMA_DEFS`, calling `parse_schema` on the type definition works, as it puts the named schema into the global list, then when subsequent reads/writes happen it uses those named schemas.

This comes up specifically within the Avro spec for the RPC, where types are defined separately from the encoding of individual message parameters/responses.

https://avro.apache.org/docs/current/spec.html#Protocol+Declaration

This change would break my code.

I am okay with changing my code, but the latest commits which remove `_named_schemas` as a parameter to `schemaless_reader/writer` leave me no easy way to specify the named types (given that the global list I am using now is gone in this change, as it is the goal of this change). (and yes,  I realize that you didn't _intend_ for me to _use_ `_named_schemas`, I was planning on bring up this discussion point anyway regarding it being "non-public"..., but it being gone leaves me in a particular bind)

To be clear, I do understand why a global list may not be desirable, so not trying to argue that a global list must stay.

So at this point I'm just speaking up for my usage before this is finalized. If there is a better way, I'm open to it.

scottbelden:
@ksunden Good points. I have not thought about how this all works with RPC as we don't currently support that natively.

I'm still playing around with the implementation and I'm not sure that the current route I'm going down is going to be the final one. In fact, an idea I had which I haven't tried yet is to make it so that when `parse_schema` is called, the `named_schemas` would be included in that result. This would be useful for me because right now I have to build that structure up as I read/write and I think might be useful for what you need.

For example:

```python
schema = {
    "name": "Weather",
    "type": "record",
    "fields": [
        {
            "name": "place1",
            "type": {
                "name": "Location",
                "type": "record",
                "fields": [{"name": "city", "type": "string"}],
            }
        },
        {
            "name": "place2",
            "type": "Location",
        }
    ],
}

parsed_schema = parse_schema(schema)
```

And then `parsed_schema` would be something like this:

```
{'__fastavro_parsed': True,
 '__named_schemas': {'Location': {'fields': [{'name': 'city',
                                              'type': 'string'}],
                                  'name': 'Location',
                                  'type': 'record'},
                     'Weather': {'fields': [{'name': 'place1',
                                             'type': {'fields': [{'name': 'city',
                                                                  'type': 'string'}],
                                                      'name': 'Location',
                                                      'type': 'record'}},
                                            {'name': 'place2',
                                             'type': 'Location'}],
                                 'name': 'Weather',
                                 'type': 'record'}},
 'fields': [{'name': 'place1',
             'type': {'fields': [{'name': 'city', 'type': 'string'}],
                      'name': 'Location',
                      'type': 'record'}},
            {'name': 'place2', 'type': 'Location'}],
 'name': 'Weather',
 'type': 'record'}
```

So I think the `__named_schemas` would essentially be the same as the `SCHEMA_DEFS` which I'm assuming you are using right now.

ksunden:
I think that would be workable (if require me to pass some things around, which is okay)

My avro-rpc implementation is [here](https://gitlab.com/yaq/yaq-python/-/tree/master/yaqd-core/yaqd_core/avrorpc). It is currently only distributed as part of a specific project, but I plan to break it out (and actually test it against other implementations, which I don't currently do... Following a subset of the spec I think, though have not yet implemented _all_ of the spec)

Is this something you would want upstreamed here, or would you prefer it is a separate package that uses fastavro, but is distributed separately? 

(the linked folder is mostly sans-IO, with the edge TCP and actual method calls a folder up, and a bit more tied to the specifics of the library it is in now, part of breaking it out would be decoupling this from my current object model, but also preferable providing an http IO layer (currently I do raw sockets with stateful RPC, http in avro is specifically stateless))

scottbelden:
@gaborbernat I don't know that I'm completely done with this yet, but I'm starting to become more okay with the state of it. I was wondering if you'd be able to use it and see if it works for you. I don't think there should be any API changes so if anything unexpected happens I would be interested in hearing that. I could potentially create a pre-release that could be pip installed if that's easier, but that might be sometime next week if I do that since I would want to take a bit more time to check over what I have so far.

@ksunden I'd also be interested to hear from you and how this might affect your situation. For the moment you can do things like `parse_schema(schema, _named_schemas=foo)`. Initially my thought was that the `_named_schemas` would be more internal since I didn't think there would be a reason people would want that, but I am leaning towards removing the underscore and having a bit more information about what that variable means and how to use it for the use-cases that need it.

gaborbernat:
Can do a quick check on Monday, thanks for the tag 👍

gaborbernat:
@scottbelden can confirm that this PR would fix our issue (at least on quick unittest validation).
appl

scottbelden:
@gaborbernat Thanks. I think I will probably continue to think over this and see if there's anything I'm overlooking. So I'll probably merge it over the weekend if I continue to be happy with it.

scottbelden:
Will probably push a new release tomorrow.

ksunden:
I was able to get this to work, though using `_named_schemas` passed to `parse_schema`, and had to explicitly pass `expand=True`, as there is no way to pass named schemas into `schemaless_reader/writer`, and without expanding, named schemas are kept as the string.

Note also: the type stub was not updated with the new signature

scottbelden:
I would be cautious about using `expand` as it can generate schemas that would later fail to parse. If you are only using schemaless reader/writer then maybe it will be okay, but if you were using the normal reader/writer then I think you would run into problems when reading.

Maybe I don't fully understand the use-case yet, but I would think you should be able to do something like this:

```python
named_schemas = {}

parsed1 = parse_schema(schema1, _named_schemas=named_schemas)
parsed2 = parse_schema(schema2, _named_schemas=named_schemas)
parsed3 = parse_schema(schema3, _named_schemas=named_schemas)

schemaless_writer(fo, parsed3, record)
```

ksunden:
That does not work, as `parse_schema("named_schema", _named_schema={"named_schema":{"type":"record", "name":"named_schema", "fields":[]}})` returns the string `"named_schema"` right back, which then is not parsed when handed to `schemaless_writer`

scottbelden:
Ah, I see. Unfortunately I'm not sure that I have any other solutions at the moment.

PR #453, opened 2020-07-29
Check actual decimal value precision during serialization

### LINKED ISSUE #452: Schema's decimal precision is ignored when serializing a decimal
A decimal number is not checked for actual precision before serializing. This leads to loss of significant digits during deserialization.

A check similar to decimal scale validation is needed.

```
from decimal import Decimal, getcontext
from io import BytesIO

import fastavro

getcontext().prec = 18


def serialize(schema, data):
    bytes_writer = BytesIO()
    fastavro.schemaless_writer(bytes_writer, schema, data)
    return bytes_writer.getvalue()


def deserialize(schema, binary):
    bytes_writer = BytesIO()
    bytes_writer.write(binary)
    bytes_writer.seek(0)

    res = fastavro.schemaless_reader(bytes_writer, schema)
    return res


def test_ignored_precision_value():
    schema = {
        'name': 'test_ignored_precision_value',
        'type': 'bytes',
        'logicalType': 'decimal',
        'precision': 4,
        'scale': 2
    }

    data1 = Decimal('123456789012345.1')
    print(f'Original: {data1}')
    binary = serialize(schema, data1)
    data2 = deserialize(schema, binary)
    print(f'Restored: {data2}')
    assert (data1 == data2)


if __name__ == '__main__':
    test_ignored_precision_value()
```

matpuk:
Fixes #452

### DIFF
--- a/fastavro/_logical_writers.pyx
+++ b/fastavro/_logical_writers.pyx
@@ -121,8 +121,13 @@ cpdef prepare_bytes_decimal(object data, schema):
     if not isinstance(data, decimal.Decimal):
         return data
     scale = schema.get('scale', 0)
+    precision = schema['precision']
 
     sign, digits, exp = data.as_tuple()
+
+    if len(digits) > precision:
+        raise ValueError(
+            'The decimal precision is bigger than allowed by schema')
 
     delta = exp + scale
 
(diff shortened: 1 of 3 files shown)


matpuk:
What is wrong with codecov/project? How to fix it?

scottbelden:
It acts up sometimes. Thanks for the PR. I'm planning to merge this in but I will probably wait until I merge another PR that is just about done before releasing.

PR #455, opened 2020-07-31
Add array.array to array, bytearray to bytes and write_int improvement

fthyssen:
First of all, sorry for adding too many items in one PR.  They're well separated in commits, so I can easily drop one or more parts.
As a second remark, I should probably add I have no experience with cython, so I'll be happy to read your comments.

The first change allows for python's `array.array` to be serialized to an array.
The second change allows for a `bytearray` to be serialized to bytes.  I had to move some logic to `write_utf8` to avoid performance loss.
The last commit introduces a small change in `write_int`, that improves cython performance for me.

I also added a `timeit`-based benchmark, to easily follow up the effects:

original:
```
Array from list                   100 records, best of   1000
========================================================
write fastavro               100 records, best of   1000:  14.400871 ms
read  fastavro               100 records, best of   1000:  15.655241 ms
write schemaless fastavro    100 records, best of   1000:  14.669046 ms
read  schemaless fastavro    100 records, best of   1000:  13.781189 ms
validate fastavro            100 records, best of   1000:  15.007082 ms

real    1m16.688s
user    1m15.806s
sys     0m0.705s
```

after array.array and bytearray addition:
```
Array from list                   100 records, best of   1000
========================================================
write fastavro               100 records, best of   1000:  14.400178 ms
read  fastavro               100 records, best of   1000:  15.634401 ms
write schemaless fastavro    100 records, best of   1000:  14.759453 ms
read  schemaless fastavro    100 records, best of   1000:  13.944073 ms
validate fastavro            100 records, best of   1000:  15.596337 ms

real    1m20.245s
user    1m19.191s
sys     0m0.752s
```

after write_int change:
```
Array from list              100 records, best of   1000
========================================================
write fastavro               100 records, best of   1000:  11.314640 ms
read  fastavro               100 records, best of   1000:  15.760850 ms
write schemaless fastavro    100 records, best of   1000:  11.621763 ms
read  schemaless fastavro    100 records, best of   1000:  13.935867 ms
validate fastavro            100 records, best of   1000:  15.673204 ms

real    1m11.119s
user    1m10.324s
sys     0m0.626s
```

array.array vs list:
```
Array from list              100 records, best of    100
========================================================
write fastavro               100 records, best of    100:  11.315835 ms
read  fastavro               100 records, best of    100:  15.406354 ms
write schemaless fastavro    100 records, best of    100:  11.669317 ms
read  schemaless fastavro    100 records, best of    100:  14.000947 ms
validate fastavro            100 records, best of    100:  15.810282 ms

Array from array             100 records, best of    100
========================================================
write fastavro               100 records, best of    100:   9.927711 ms
read  fastavro               100 records, best of    100:  15.353774 ms
write schemaless fastavro    100 records, best of    100:  10.219035 ms
read  schemaless fastavro    100 records, best of    100:  14.010455 ms
validate fastavro            100 records, best of    100:  17.267594 ms
```

Validation for python arrays is slower than for lists, and could be improved for specific cases based on `array.itemsize` and signedness and/or cython memoryviews.

Finally, `write_python_array` probably could be improved by also adding cases where for example `float` isn't the exact type, but part of a union.

### DIFF
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -114,7 +114,7 @@ cdef inline write_double(bytearray fo, double datum):
     fo += ch_temp[:8]
 
 
-cdef inline write_bytes(bytearray fo, bytes datum):
+cdef inline write_bytes(bytearray fo, const unsigned char[:] datum):
     """Bytes are encoded as a long followed by that many bytes of data."""
     write_long(fo, len(datum))
     fo += datum
@@ -123,7 +123,9 @@ cdef inline write_bytes(bytearray fo, bytes datum):
 cdef inline write_utf8(bytearray fo, datum):
     """A string is encoded as a long followed by that many bytes of UTF-8
     encoded character data."""
-    write_bytes(fo, utob(datum))
+    b_datum = utob(datum)
+    write_long(fo, len(b_datum))
+    fo += b_datum
 
 
 cdef inline write_crc32(bytearray fo, bytes bytes):
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -55,7 +55,7 @@ cdef inline write_int(bytearray fo, datum):
     """
     cdef ulong64 n
     cdef unsigned char ch_temp[1]
-    n = (datum << 1) ^ (datum >> 63)
+    n = ~(datum << 1) if datum < 0 else (datum << 1)
     while (n & ~0x7F) != 0:
         ch_temp[0] = (n & 0x7f) | 0x80
         fo += ch_temp[:1]
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -6,6 +6,8 @@
 # http://svn.apache.org/viewvc/avro/trunk/lang/py/src/avro/ which is under
 # Apache 2.0 license (http://www.apache.org/licenses/LICENSE-2.0)
 
+from cpython cimport array
+import array
 import json
 from binascii import crc32
 from os import urandom
@@ -165,6 +167,27 @@ cdef write_array(bytearray fo, list datum, schema, dict named_schemas):
         dtype = schema['items']
         for item in datum:
             write_data(fo, item, dtype, named_schemas)
+    write_long(fo, 0)
+
+
+cdef write_python_array(bytearray fo, array.array datum, schema, named_schemas):
+    """Array specialization for python arrays."""
+    if len(datum) > 0:
+        write_long(fo, len(datum))
+        record_type = extract_record_type(schema['items'])
+        if record_type in ('int', 'long'):
+            for idx in range(len(datum)):
+                write_int(fo, datum[idx])
+        elif record_type == 'float':
+            for idx in range(len(datum)):
+                write_float(fo, datum[idx])
+        elif record_type == 'double':
+            for idx in range(len(datum)):
+                write_double(fo, datum[idx])
+        else:
+            dtype = schema['items']
+            for item in datum:
+                write_data(fo, item, dtype, named_schemas)
     write_long(fo, 0)
 
 
@@ -330,7 +353,9 @@ cpdef write_data(bytearray fo, datum, schema, dict named_schemas):
     elif record_type == 'enum':
         return write_enum(fo, datum, schema, named_schemas)
     elif record_type == 'array':
-        if not isinstance(datum, list):
+        if isinstance(datum, array.array):
+            return write_python_array(fo, datum, schema, named_schemas)
+        elif not isinstance(datum, list):
             datum = list(datum)
         return write_array(fo, datum, schema, named_schemas)
     elif record_type == 'map':
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -50,7 +50,7 @@ cdef inline write_boolean(bytearray fo, bint datum):
     fo += ch_temp[:1]
 
 
-cdef inline write_int(bytearray fo, long long datum):
+cdef inline write_int(bytearray fo, long64 datum):
     """int and long values are written using variable-length, zig-zag coding.
     """
     cdef ulong64 n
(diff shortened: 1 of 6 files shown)


scottbelden:
Thanks for this. I'll try to take a look soon.

fastavro/_write.pyx
```diff
@@ -112,7 +114,7 @@ cdef inline write_double(bytearray fo, double datum):
     fo += ch_temp[:8]
 
 
-cdef inline write_bytes(bytearray fo, bytes datum):
+cdef inline write_bytes(bytearray fo, const unsigned char[:] datum):
```
scottbelden, line 117: I'm not too familiar with all the different types and how cython uses them. Is there a noticeable impact from this change?

fastavro/_write.pyx
```diff
@@ -121,7 +123,9 @@ cdef inline write_bytes(bytearray fo, bytes datum):
 cdef inline write_utf8(bytearray fo, datum):
     """A string is encoded as a long followed by that many bytes of UTF-8
     encoded character data."""
-    write_bytes(fo, utob(datum))
+    b_datum = utob(datum)
```
scottbelden, line 126: I see that this would save a function call, but I would assume that the function call wouldn't be as expensive in cython as it is in python. Is there a measurable difference here?

fastavro/_write.pyx
```diff
@@ -168,6 +172,31 @@ cdef write_array(bytearray fo, list datum, schema, dict named_schemas, fname):
...
+                        array.array datum,
+                        schema,
+                        named_schemas,
+                        fname):
+    """Array specialization for python arrays."""
+    if len(datum) > 0:
+        write_long(fo, len(datum))
+        record_type = extract_record_type(schema['items'])
+        if record_type in ('int', 'long'):
+            for idx in range(len(datum)):
```
scottbelden, line 185: Can these loops be changed to the following?

```python
for item in datum:
    write_X(fo, item)
```

Or does that change the performance?

fastavro/_write.pyx
```diff
@@ -53,7 +55,7 @@ cdef inline write_int(bytearray fo, datum):
     """
     cdef ulong64 n
     cdef unsigned char ch_temp[1]
-    n = (datum << 1) ^ (datum >> 63)
+    n = ~(datum << 1) if datum < 0 else (datum << 1)
```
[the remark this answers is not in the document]
fthyssen: In the tables above, it was the change from 14.4ms to 11.3ms.  It's equivalent, and clearer if we go to the original protocol-buffer [reference](https://developers.google.com/protocol-buffers/docs/encoding?csw=1#types) in the spec that says

> `(n << 1) ^ (n >> 63)`
> Note that the second shift – the (n >> 31) part – is an arithmetic shift. So, in other words, the result of the shift is either a number that is all zero bits (if n is positive) or all one bits (if n is negative).

In other words
- xor with all zeros if it's positive, so do nothing, or
- xor with all ones if it's negative, so a not-operator.

I'm assuming cython does two pythonic shifts, which appears to be more expensive than one comparison and one shift.

That said, I committed a version that keeps the original formula, but enforces the datum to be a long.  This makes it even more efficient (~9ms for that test), but I was reluctant to do so as I wasn't sure about python 2.7.

I also added a test to verify the extreme values, and compared encoding offline with avro by adding these lines to `timeit_benchmark`:
```
    list_long_record = [-2, -1, 0, 1, 2, 2147483647, -2147483648,
                        9223372036854775807, -9223372036854775808]

    test_case("Compare long encoding", array_schema,
              list_long_record, 1, 1, run_avro=args.run_avro)
```

fastavro/_write.pyx
fthyssen: The line-score for
```for entry in datum```
reduces from 43 to 5, but the line-score for
```write_int(fo, entry)```
increases from 1 to 4, so there's probably a tipping-pont somewhere for larger arrays.
When I test it roughly, the difference is below 2%, with the advantage tipping somewhere between 50 and 200 array-size, so I'll go with the easiest to read.

File: fastavro/_write.pyx (hunk 1 above)
Comment (fthyssen) [reply fastavro/_write.pyx:117]: This is what makes it possible to use both bytearrays and bytes.
It is also recommended [here](https://cython.readthedocs.io/en/latest/src/tutorial/strings.html#accepting-strings-from-python-code)

> In the case that the API only deals with byte strings, i.e. binary data or encoded text, it is best not to type the input argument as something like bytes, because that would restrict the allowed input to exactly that type and exclude both subtypes and other kinds of byte containers, e.g. bytearray objects or memory views.

You do lose out on performance, which is why I removed the function call from `write_utf8`.

File: fastavro/_write.pyx (hunk 2 above)
Comment (fthyssen) [reply fastavro/_write.pyx:126]: Here to make up for the performance loss from the changed signature of write_bytes.  I'm guessing as a result of casting away from the known bytes-type here.

File: fastavro/_write.pyx [hunk 5]
```diff
@@ -48,7 +50,7 @@ cdef inline write_boolean(bytearray fo, bint datum):
     fo += ch_temp[:1]
 
 
-cdef inline write_int(bytearray fo, datum):
+cdef inline write_int(bytearray fo, long long datum):
```
fthyssen, line 53: I just realized it's probably preferred to use the typename convention:

```suggestion
cdef inline write_int(bytearray fo, long64 datum):
```

scottbelden: Agreed

tests/test_types.py
```diff
@@ -92,3 +92,24 @@ def test_int_in_string_raises():
     # Raises AttributeError on py2 and TypeError on py3
     with pytest.raises((TypeError, AttributeError)):
         serialize(schema, *records)
+
+
+def test_int_bytes():
+    test_cases = [(0, b"\x00"),
```
scottbelden, line 98: Can you change this to use the `pytest.mark.parametrize`? An example is here:

https://github.com/fastavro/fastavro/blob/57fbe15742eb03e3a19882ebcf16b1adb2ea32dc/tests/test_logical_types.py#L196-L226

scottbelden:
Thanks! I'll probably get this out in a release today or tomorrow.

PR #459, opened 2020-08-05
fix timestamp-millis for problem cases

scottbelden:
https://github.com/fastavro/fastavro/issues/458


scottbelden:
I should also look into `timestamp-micros` before merging as I'm sure it probably has the same problems.

PR #461, opened 2020-08-09
Improved decimal schema validation

### LINKED ISSUE #457: Improve decimal schema validation
Currently, I found 3 issues with decimal schema:

1. Negative values for precision and scale are allowed
2. There is no check that specified precision will fit into fixed number of bytes
3. There is no check that precision >= scale

Is it ok to fix all of the issues in one PR?

matpuk:
Resolves #457:

- Added missing check for positiveness of scale and precision
- Added check for scale <= precision
- Added check for fixed type dicimal precision limit

### DIFF
--- a/fastavro/_schema_py.py
+++ b/fastavro/_schema_py.py
@@ -1,5 +1,5 @@
 # cython: auto_cpdef=True
-
+import math
 from os import path
 from copy import deepcopy
 import json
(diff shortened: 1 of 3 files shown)


fastavro/_schema_py.py
```diff
@@ -267,16 +267,36 @@ def _parse_schema(
...
+                if not isinstance(precision, int) or precision <= 0:
+                    raise SchemaParseException(
+                        "decimal precision must be a postive integer, "
+                        + "not {}".format(precision)
+                    )
+                if schema_type == "fixed":
+                    # https://avro.apache.org/docs/current/spec.html#Decimal
+                    size = schema["size"]
+                    max_precision = int(
+                        math.floor(math.log10(2) * (8 * size - 1))
```
scottbelden, line 287: I see that this is the same formula that used in the apache avro libraries, but isn't it ever so slightly different than the formula on the spec page? The spec page has an extra `- 1` near the end. In reality it doesn't seem to matter because we are doing a floor function and the `size` can only be whole integers. I'm guessing the final `- 1` is dropped to simplify the function.

Or am I misreading the function in the spec?

matpuk: Yes, you are right. The -1 part is dropped because in the correct full formula: log10(2^(8*n-1)) + log10(1-1/2^(8*n-1)), the last part will go to 0 with bigger 'n'.

matpuk: And, btw, I took this simplified formula from reference official avro C++ library.

PR #466, opened 2020-08-24
Set python_requires='>=3.5'

Nicusor97:
From your changelog seems like in the latest version you dropped Python 2 support but it is incomplete.  You need to set `python_requires` in your setup.py that will prevent users with Python 2.7 from downloading a sdist version they cannot build.

Note: supporting python_requires requires setuptools>=24.2.0 and pip>=9.0.0 to benefit from it.
Details here: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,7 @@ setup(
         'Topic :: Software Development :: Libraries',
         'Topic :: Scientific/Engineering :: Information Analysis',
     ],
-    python_requires='>=3.5',
+    python_requires='>=3.0',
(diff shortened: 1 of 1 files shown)


Nicusor97:
Here is the current issue if you try to install latest version of fastavro on Python 2: 
```         
if sys.implementation.name != "pypy":
     AttributeError: 'module' object has no attribute 'implementation'
```
The `sys` module has the `implementation` attribute only on Python 3.

Issue #505, opened 2021-01-30
Compile schemas into executable code objects


spenczar:
Warning: slightly crazy ideas ahead.

In the course of working through the hand-written decoder in #503, I got to thinking about the structure of fastavro's parsing. When it is reading a deep union or record, it needs to do a bunch of bookkeeping and dictionary lookups and function calls and stuff to make sense of things. If you were writing the deserialization function by hand for a specific schema, that's not how you'd do it; you'd directly call `read_utf8` and then `read_long` and then `read_long` and then `read_bool` or whatever - you'd know just the right order to call them and wouldn't need to look anything up.

But of course, you can't do that for the generic case of a deserializing library that doesn't know the schema ahead of time.

Or can you? Python has the `compile` and `eval` builtins. You _could_ parse the schema into a big string which is the Python code you would have hand-written. You could compile and evaluate that string, and then invoke the resulting code object to actually do parsing. You only need to compile the schema once, and then invoke the code object each time you want to parse with that schema.

I have a proof of concept of this. For my weird astronomical alerts, it is about 20-30% faster than fastavro's  `reader`. 

The compiler POC is here: [schema_compile.py](https://github.com/ZwickyTransientFacility/alertbase/blob/322f8a67dc6cff68e553c25ad1f3bc1b688199bd/src/alertbase/schema_compile.py). For debugging, I printed out the code object it generates, which you can find [here](https://github.com/ZwickyTransientFacility/alertbase/blob/322f8a67dc6cff68e553c25ad1f3bc1b688199bd/gen.py) - it's nasty but it works.

Using the same benchmarks as #503, I ran some numbers:

method | Single-file time (ms) | Tarball time (per object) (ms) | Tarball MB/sec |
| ------ | ----- | ----- | ----- |
compiled | 0.661| 1.210 | 54.84 |
fastavro | 0.903 | 1.546 | 42.90 |
avro | 5.883 | 6.914 | 9.59 |

(I've omitted the subschema and "manual decoder" cases because they're not relevant here; the compiled schema parses _all_ fields).

My code is not perfect by any means - it's just a hacky proof of the idea: 
- I didn't implement `fixed` or `map` types, because they aren't in my schema.
- Lots of the generated code is dead code; there is a lot that could be optimized away.
- I don't think I handle some weird cases right, like arrays-of-arrays. I definitely don't handle recursion right.
- I do no error handling or sanity checks or anything, so corrupted data would probably go haywire.

[...]

PR #508, opened 2021-02-11
(work-in-progress, do not merge): Compile schemas into executable code

spenczar:
We talked about this in #505.

I have made some good progress, which I'd like to share. I generate a Python AST for a given schema, which means I don't need to use `eval` which is nice. It compiles into a function which can then read data from a file-like source and spits out the deserialized data.

In early microbenchmarks, this is performing very well. Benchmark-running code is in `tests/benchmark_compilation.py`; the benchmark test cases are in `tests/test_ast_compilation.py`. Full results are [in a gist](https://gist.github.com/spenczar/ee835e213ddd3a04ddcd80639e463aa6), but in summary:
 - Reading records and unions is ~3-4x faster
 - Reading maps and arrays are ~1.2x faster
 - Reading enums, fixed types, logical types, and schemas which are just a primitive is ~2x faster

Of course, this is balanced by the compilation time, which can be a few milliseconds even for a simple schema. It nice to provide a way to cache compilation (in case they're reading a lot of files, all of which have the same schema, for example).

By hacking around with the `linecache` package, I'm able to get generated source code to appear in tracebacks, which is nice. Still, the generated code can be pretty strange.

This sort of code generation is not easy to read, and might be so hard to maintain that you don't want to pursue this - I'd understand. I have become very very familiar with the `ast` package by this point, but it's weird stuff. I am also not so sure how well Python versions will maintain compatibility going forward.

TODOs
--------------
- [x] Compile reader for a given schema
- [ ] Compile writer for a given schema
- [x] Handle recursive schemas
- [ ] Handle schema evolution
   - [ ] Skip fields which are removed in reader schema
   - [ ] Set defaults for fields which are added in reader schema
   - [ ] Handle type promotion
   - [ ] Handle field reordering
 - [x] Remove dependencies on Python 3.9
 - [ ] Use in existing fastavro tests


spenczar:
An example of generated code, if you use `ast.unparse` to decompile the AST into source code (`ast.parse` as added in Python 3.9): https://gist.github.com/spenczar/dcf2828958e13edcd6a2c53091bc4c0b

scottbelden:
This is really interesting and the initial benchmarks do look nice. I've worked with `ast` in the past so I'm at least a little familiar with the concept and format.I'll try to find some time to take a deeper look at this and give some more detailed thoughts.

[I also played around with an approach similar to the `spavro`](https://github.com/fastavro/fastavro/pull/507) idea of pre-parsing the schema and generating a writer function that only depends on the data, but I haven't figured out if it is actually beneficial.

scottbelden:
@spenczar Hey, I wanted to take an initial stab at trying to replicate some of the results you've gotten, but I think I might be doing something wrong...

I found the schema you mentioned in your gist, but I didn't have any data to work with, so I tried generating some arbitrary data. Then I tried to set up a script similar to `test_ast_compiler_read_file` in `tests/test_ast_compile.py`. Here's a gist with the schema you provided, the test records I generated, and the script I used to do the timing: https://gist.github.com/scottbelden/9501ad51cb4892a2ab32789d3777a613

At the end I got these timings:

```
Write time: 0.45223498344421387
Read time: 0.19674086570739746
Compile Read time: 0.17552995681762695
```

I see that the compiled read time is indeed faster. Does this performance change match what you have been seeing and what you mentioned in the original post, or am I doing something wrong? Thanks!

spenczar:
Nice, thanks for the concise speed test. Yes, I get similar results:
```
-> % python3.9 test_speed.py             
Write time: 0.3856966495513916
Read time: 0.17666053771972656
Compile Read time: 0.1411588191986084
```

So about 20% faster.

The dramatic benchmarks (like "70% faster") come from micro benchmarks, which you can run with `python tests/benchmark_compilation.py` in this branch. Those might not be meaningful.

I was seeing better results when working a bit with real data - more like 30%. I'll try to get some of that real data into a gist so we can run your script against it - one sec.

spenczar:
Okay, raw data here: https://github.com/spenczar/astro_data_share/blob/main/astro_records.avro. Couldn't do a gist because it's too big, and couldn't do JSON because it's too big, sorry.

It's 1000 records created with `fastavro.writer`. It's pretty much the same performance (relatively) that you saw in your synthetic test:

```
1000
Write time: 0.9514408111572266
Read time: 0.36310553550720215
Compile Read time: 0.2846710681915283
```
This is a 21% improvement in runtime.
---
So - is this worth all the complexity of this change? I'm not totally convinced, myself. The microbenchmarks, like of very simple records, were way more convincing.

spenczar:
I have spun this off into a separate project, https://github.com/spenczar/avroc. Thanks for thinking it through with me and exploring the design - I really appreciate it.

PR #523, opened 2021-03-18
Preserve precision of double-point floats which are encoded in a union

spenczar:
Fixes https://github.com/fastavro/fastavro/issues/437

I think there's not going to be much better than this: we can check whether "double" is one of the other options in a union when we're considering encoding as a "float." If not, fine, we'll stick with float. But if "double" is an option, we should take it.


tests/test_fastavro.py
```diff
@@ -2649,3 +2649,14 @@ def test_schema_migration_should_fail_with_different_names():
 
     with pytest.raises(SchemaResolutionError):
         roundtrip(writer_schema, records, reader_schema=reader_schema)
+
+
+def test_union_of_float_and_double_keeps_precision():
+    """https://github.com/fastavro/fastavro/issues/437"""
+    schema = ["float", "double"]
```
scottbelden, line 2656: Could you change this line to the following so that it covers the case where we have found a `float` and we are continuing to look through the schemas but find something other than double?

```suggestion
    schema = ["float", "string", "double"]
```

tests/test_fastavro.py
```diff
@@ -2649,3 +2649,14 @@ def test_schema_migration_should_fail_with_different_names():
 
     with pytest.raises(SchemaResolutionError):
         roundtrip(writer_schema, records, reader_schema=reader_schema)
+
+
+def test_union_of_float_and_double_keeps_precision():
```
scottbelden, line 2654: Could we also add a test where there is a float but no double. For example, a schema like `schema = ["float", "double"]`

scottbelden (changes requested):
This looks good to me. Just a few small suggestions. Thanks!

scottbelden (approved):

PR #530, opened 2021-04-06
Apply logical type preparation when checking for best union type

ksunden:
Previously validate would pass (and it does the prepare step) but then the datum was still pre preparation when it tried to do anything further. 
As far as I can tell this only really matters for record types, but made more general in case there was an edge case I was missing. 

The provided test shows this for an ndarray record type which would fail because set(ndarray) does not work [well, it actually does for 1D arrays, but not 2+ dimensional arrays... even in the 1d case it was doing something funky that technically worked, but only on accident]

### DIFF
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -295,15 +295,14 @@ cdef write_union(bytearray fo, datum, schema, dict named_schemas, fname):
                     continue
 
             if _validate(datum, candidate, named_schemas, raise_errors=False):
-                if isinstance(candidate, dict):
+                record_type = extract_record_type(candidate)
+                if record_type == "record":
                     logical_type = extract_logical_type(candidate)
                     if logical_type:
                         prepare = LOGICAL_WRITERS.get(logical_type)
                         if prepare:
                             datum = prepare(datum, candidate)
 
-                record_type = extract_record_type(candidate)
-                if record_type == "record":
                     candidate_fields = set(
                         f["name"] for f in candidate["fields"]
                     )
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -295,6 +295,13 @@ cdef write_union(bytearray fo, datum, schema, dict named_schemas, fname):
                     continue
 
             if _validate(datum, candidate, named_schemas, raise_errors=False):
+                if isinstance(candidate, dict):
+                    logical_type = extract_logical_type(candidate)
+                    if logical_type:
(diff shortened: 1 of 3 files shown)


scottbelden:
This is an interesting use case of a `logicalType`. None of the logical types currently defined in the Avro specification use `record` as the base type. Most use a primitive type, but `duration` uses `fixed` which is also a named type. The specification states that `A logical type is an Avro primitive or complex type with extra attributes to represent a derived type.` so I don't think there is any reason to say that a record can't be a logical type.

I think something like this makes sense, but I'm going to think about it a little and get back to you.

fastavro/_write.pyx
```diff
@@ -295,6 +295,13 @@ cdef write_union(bytearray fo, datum, schema, dict named_schemas, fname):
                     continue
 
             if _validate(datum, candidate, named_schemas, raise_errors=False):
+                if isinstance(candidate, dict):
```
scottbelden, line 298: I saw the part about the check being done here for the general case, but I think we can safely keep this logic in the specific record case. Therefore, could you move this down under the `if record_type == "record":` block? We can also get rid of the `if isinstance(candidate, dict):` check at that point.

fastavro/_write_py.py
```diff
@@ -183,6 +183,13 @@ def write_union(encoder, datum, schema, named_schemas, fname):
                     continue
 
             if _validate(datum, candidate, named_schemas, raise_errors=False):
+                if isinstance(candidate, dict):
```
scottbelden, line 186: Same comment as for the other `write` file.

tests/test_logical_types.py
```diff
@@ -474,3 +475,80 @@ def test_local_timestamp_micros():
...
+
+
+class Interface:
+    def __init__(self, array_interface):
+        array_interface["shape"] = tuple(array_interface["shape"])
+        self.__array_interface__ = array_interface
+
+
+def read_ndarray(data, writer_schema, reader_schema):
+    import numpy as np
```
scottbelden, line 487: This can be removed since the import is at the top of the file.

scottbelden (changes requested):
I think this makes sense. Just a few small comments and then it should be good to go. Thanks!

tests/test_logical_types.py
ksunden: ah, this was copy pasted from some code that I wanted to register the logical type conversion, but not import numpy until it actually needed it

scottbelden (approved):

PR #541, opened 2021-05-09
538 fix schema expansion of parsed schemas

### LINKED ISSUE #538: schema.expand_schema fails to consider __named_schemas member
# Summary

When using `schema.expand_schema` on a schema that was created using `parse_schema(some_dict, some_named_schemas)` it does not consider the `__named_schemas` member on `parse_schema`'s output dict.
This leads to the unexpected behaviour that the schema produced by `parse_schema` has a valid named schema for some key `A` but trying to expand the schema by running `schema.expand_schema(parse_schema(some_dict, some_named_schemas))` throws `UnknownType(A)`.

# Example

Consider the following example:

```
from fastavro.schema import expand_schema
from fastavro import parse_schema

referenced = {
    "com.namespace.dependencies.Dependency": {
        "name": "Dependency",
        "namespace": "com.namespace.dependencies",
        "type": "record",
        "fields": [
            {"name": "sub_field_1", "type": "string"}
        ]
    }
}

original_schema = {
    "name": "MasterSchema",
    "namespace": "com.namespace.master",
    "type": "record",
    "fields": [{
        "name": "field_2",
        "type": "com.namespace.dependencies.Dependency"
    }]
}

schema_works = parse_schema(original_schema, named_schemas=referenced, expand=True) # works
schema_breaks = expand_schema(parse_schema(original_schema, named_schemas=referenced)) # throws UnknownType

```

# Potential cause

I think this behaviour occurs because `parse_schema` is called from `expand_schema` without checking if its input `dict` has a `__named_schemas` member and not passing it on to `parse_schema` explicitly ([_schema.py#L158](https://github.com/fastavro/fastavro/blob/c52dbfd35d7b26945a24d2225f8db3827866170b/fastavro/_schema_py.py#L158)) _and_ the local `named_schemas` dictionary in the `parse_schema` function only being populated from the `__named_schemas` member _after_ the `if expand` check ([_schema.py#L224ff.](https://github.com/fastavro/fastavro/blob/c52dbfd35d7b26945a24d2225f8db3827866170b/fastavro/_schema_py.py#L224)).

# Workaround

Ofc I didn't read the docs properly, otherwise I wouldn't have attempted this at all because what _does work_ to achieve the same goal is simply using `parse_schema` and setting the named parameter `expand` to `True`.

I still think that the behaviour of `expand_schema` on a `dict` with a `__named_schemas` member is unexpected. Do you agree? If so I think I could provide a PR for this.

srfc:
Fixes #538

### DIFF
--- a/fastavro/_schema_py.py
+++ b/fastavro/_schema_py.py
@@ -220,6 +220,10 @@ def parse_schema(
     if named_schemas is None:
         named_schemas = {}
 
+    if isinstance(schema, dict) and "__fastavro_parsed" in schema:
+        for key, value in schema["__named_schemas"].items():
+            named_schemas[key] = value
+
     if _force or expand:
         return _parse_schema(schema, "", expand, _write_hint, set(), named_schemas)
     elif isinstance(schema, dict) and "__fastavro_parsed" in schema:
--- a/fastavro/_schema_py.py
+++ b/fastavro/_schema_py.py
@@ -227,8 +227,6 @@ def parse_schema(
     if _force or expand:
         return _parse_schema(schema, "", expand, _write_hint, set(), named_schemas)
     elif isinstance(schema, dict) and "__fastavro_parsed" in schema:
-        for key, value in schema["__named_schemas"].items():
-            named_schemas[key] = value
         return schema
     elif isinstance(schema, list):
         # If we are given a list we should make sure that the immediate sub
--- a/fastavro/_schema.pyx
+++ b/fastavro/_schema.pyx
@@ -74,6 +74,10 @@ def parse_schema(
     if named_schemas is None:
         named_schemas = {}
 
+    if isinstance(schema, dict) and "__fastavro_parsed" in schema:
+        for key, value in schema["__named_schemas"].items():
+            named_schemas[key] = value
+
     if _force or expand:
         return _parse_schema(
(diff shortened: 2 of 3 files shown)


fastavro/_schema_py.py
```diff
@@ -220,11 +220,13 @@ def parse_schema(
     if named_schemas is None:
         named_schemas = {}
 
+    if isinstance(schema, dict) and "__fastavro_parsed" in schema:
+        for key, value in schema["__named_schemas"].items():
+            named_schemas[key] = value
+
```
srfc, line 226: @scottbelden As far as I can tell, the `_py.py` fallback implementation is currently not covered altogether resulting in changes (or additions) in them to always decrease code cov. Do you have any advice on how to handle this?

scottbelden: It looks like the tests needed to be approved to run. I don't remember that being the case in the past but maybe github made some changes so that forks needs approval to run.

scottbelden:
Thanks! Merging.

fastavro/_schema_py.py
srfc: Ah, alright. I was a bit confused, thanks for the explanation.

srfc:
> Thanks! Merging.

Happy for that, have a nice remaining Monday!

PR #562, opened 2021-10-05
support python 3.10

scottbelden:


scottbelden:
Will need to wait for wheels of dependencies like numpy.

PR #579, opened 2022-02-05
Add field_name to the READERS functions

cchantep:
When extending the `READERS` functions (`READERS['union'] = custom_union_reader`), it can be very useful to be given the name of the field.

### DIFF
--- a/fastavro/_read_py.py
+++ b/fastavro/_read_py.py
@@ -139,6 +139,7 @@ def read_null(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     return decoder.read_null()
 
@@ -153,6 +154,7 @@ def read_boolean(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     return decoder.read_boolean()
 
@@ -227,6 +229,7 @@ def read_int(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     return decoder.read_int()
 
@@ -241,6 +244,7 @@ def read_long(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     return decoder.read_long()
 
@@ -255,6 +259,7 @@ def read_float(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     return decoder.read_float()
 
@@ -269,6 +274,7 @@ def read_double(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     return decoder.read_double()
 
@@ -283,6 +289,7 @@ def read_bytes(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     return decoder.read_bytes()
 
@@ -297,6 +304,7 @@ def read_utf8(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     return decoder.read_utf8()
 
@@ -311,6 +319,7 @@ def read_fixed(
     named_schemas=None,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     size = writer_schema["size"]
     return decoder.read_fixed(size)
@@ -327,6 +336,7 @@ def read_enum(
     named_schemas,
     reader_schema=None,
     return_record_name=False,
+    field_name=None,
 ):
     symbol = writer_schema["symbols"][decoder.read_enum()]
     if reader_schema and symbol not in reader_schema["symbols"]:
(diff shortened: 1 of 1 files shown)


scottbelden:
I don't think I understand the use case. Can you give a larger example of how this would be used and what it does?

cchantep:
> I don't think I understand the use case. Can you give a larger example of how this would be used and what it does?

Generally speaking, if you want to extend any reader behaviour for a specific field (and any way, without knowing the field name, it's not possible to properly get the field value from `_current`).

scottbelden:
Can you post a simple script that would show how this feature would be used? I'm still not understanding... Are you talking about the `_current` variable in the json decoder?

cchantep:
> Can you post a simple script that would show how this feature would be used? I'm still not understanding... Are you talking about the `_current` variable in the json decoder?

Yes; There is no way to get the current field data in custom `decode_union` without knowing the `field_name`.

scottbelden:
I still need some example script to see how this would be used as I'm still pretty confused.

cchantep:
Aside from that I can quite easily imagine that nowing the field name is useful to implement a custom reader for some specific field ... If you want to get the current field value in such reader, you've to be able to call `decoder._current[field_name]` ... Working with index is not an option (schema field order and data/runtime order is not guaranteed to be the same).

cchantep:
`_current` may be undocumented, but that's no other way to get the value of the current field in the reader functions.

scottbelden:
Okay, so I think I finally understand.

I'm assuming you have a schema like this:

```python
schema = {
    "type": "record",
    "name": "Test",
    "namespace": "test",
    "fields": [{"name": "union", "type": ["null", "int"], "default": None}],
}
```

And you want to be able to read JSON data that looks like `{}` and `{"union": 1}`. Both of these are not valid Avro JSON, but I'm guessing you want to do some patching on the fly to make this work.

I think there might be a cleaner way to do the patch without requiring any changes to this repo. Let me think about it a bit and follow up. I feel like a change or two in `fastavro/io/parser.py` will do it. We have to parse the schema there to get it ready to be read by the JSON reader, so it feels like that's the place where the field name should be captured.

borgeser:
@scottbelden Any update on this?

You are right, we are doing some patching on the fly, using private APIs, in order to fit our needs (`optional` without the class name in the dictionary and `datetime` <-> `string` conversion).

I would be great to have more flexibility on the public API.

Another solution would be to be able to subclass `AvroJSONEncoder/Decoder` in order to implement custom behaviors.
A third one would be to be able to override `_write_py|_read_py.LOGICAL_WRITERS` and `_write_py|_read_py.WRITERS`

scottbelden:
> Another solution would be to be able to subclass `AvroJSONEncoder/Decoder` in order to implement custom behaviors.
I think this might currently work. For example, if I run this script:

```python
from io import StringIO

from fastavro import reader, writer
from fastavro.io.json_decoder import AvroJSONDecoder
from fastavro.io.json_encoder import AvroJSONEncoder


class CustomJSONDecoder(AvroJSONDecoder):
    def read_value(self, symbol):
        print("custom read")
        super().read_value(symbol)


class CustomJSONEncoder(AvroJSONEncoder):
    def write_value(self, value):
        print("custom write")
        super().write_value(value)


schema = {
    "type": "record",
    "name": "test",
    "fields": [
        {"name": "foo", "type": "string"},
        {"name": "bar", "type": "int"},
    ]
}


records = [
    {"foo": "foo", "bar": 1}
]


sio = StringIO()
writer(CustomJSONEncoder(sio), schema, records)

sio.seek(0)

for r in reader(CustomJSONDecoder(sio), schema):
    print(r)
```

It prints out:

```
custom write
custom write
custom read
custom read
{'foo': None, 'bar': None}
```

> A third one would be to be able to override `_write_py|_read_py.LOGICAL_WRITERS` and `_write_py|_read_py.WRITERS`

Creating custom logical readers/writers is supported and documented here: https://fastavro.readthedocs.io/en/latest/logical_types.html#custom-logical-types. Overriding existing logical readers/writers isn't supported or documented, but I believe should be possible via the same mechanism.

Overriding the standard readers/writers isn't supported (though could probably be done in some hacky way).

scottbelden:
I think I can fix those issues about the AttributeError. To fix that I'll probably have to make it so that `json_reader` can take in an optional decoder and `json_writer` can take in an optional encoder.

So instead you would do something like this:

```python
from fastavro import json_writer

schema = {}
records = []

with open('some-file', 'w') as out:
    json_writer(out, schema, records, encoder=CustomAvroJSONEncoder)
```

I'll see if that will fix the issue and if so I'll make a PR.

I'm not sure why the IDE is complaining. `fastavro.io.symbols` shouldn't be a package (since `symbols` is just a module).

cchantep:
Any news?

scottbelden:
Oh, sorry, I guess I forgot to mention here that the option for a custom decoder/encoder was added in https://github.com/fastavro/fastavro/pull/618. So I think you should be able to use that now.

Issue #597, opened 2022-05-17
Custom logical types not working for json_reader


arop:
Hello,

I've come across a bug on setting custom logical type handling while using `json_reader` for (de)serialization.

Here's a snippet to show it

```python
import json
import io

import fastavro


def decode(data, writer_schema=None, reader_schema=None):
    return json.loads(data)


def encode(data, *args, **kwargs):
    return json.dumps(data)


AVRO_SCHEMA = {
    "namespace": "com.issue",
    "name": "Issue",
    "type": "record",
    "fields": [
        {"name": "issue_json", "default": None, "type": {"type": "string", "logicalType": "custom_json"}},
    ],
}

ISSUE_OBJECT = {
    "issue_json": "{\"key\": \"value\"}",
}



def test_works():
    fastavro.write.LOGICAL_WRITERS['string-custom_json'] = encode
    fastavro._read_py.LOGICAL_READERS['string-custom_json'] = decode

    object_bytes = io.BytesIO(json.dumps(ISSUE_OBJECT).encode())
    re = fastavro.json_reader(fo=object_bytes, schema=AVRO_SCHEMA)
    for r in re:
        print(r)


def test_doesnt_work():
    fastavro.write.LOGICAL_WRITERS['string-custom_json'] = encode
    fastavro.read.LOGICAL_READERS['string-custom_json'] = decode

    object_bytes = io.BytesIO(json.dumps(ISSUE_OBJECT).encode())
    re = fastavro.json_reader(fo=object_bytes, schema=AVRO_SCHEMA)
    for r in re:
        print(r)


# prints: {'issue_json': {'key': 'value'}}
#test_works()

# prints {'issue_json': '{"key": "value"}'}
test_doesnt_work()
```

My guess is that `json_reader`'s code uses a different "memory" object.

Will use the working version for now, but it looks ilegal 😛

arop:
@scottbelden Taking an equivalent approach to `logical_writers` I have the code ready for PR, but I have no write permission to push a new branch...

scottbelden:
The way to do it is to make a fork of the repo, push a branch, and then submit a PR back to this repo. If you do that I would check the box to allow edits by maintainers so that I can make tweaks as necessary. Here's an example of the process: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork

PR #598, opened 2022-05-18
Fix logical readers reference

arop:
- Fixes issue #597 
- LOGICAL READERS reference wouldn't work for `json_readers` for custom logical types 
- This PR uses the same approach of `logical_writers` for the `logical_readers`


scottbelden:
@arop Thanks for putting this together! I'll try to get new release out shortly.

Issue #625, opened 2022-07-12
tuple notation even for simple unions with return_record_name


untereiner:
Hi,

I am testing a basic example with a  nullable field (ie. The error field can be an ErrorInfo or null)

```
schema = {
    "type": "record",
    "name": "Inbox",
    "fields": [
        {
            "type": "int",
            "name": "id"
        },
        {
            "type": ["null",
                     "string",
                    ],
            "name": "msg_title"
        },
        {
            "type": [
                "null",
                {
                    "type": "record",
                    "name": "ErrorInfo",
                    "fields": [
                        {
                            "name": "message",
                            "type": "string"
                        },
                        {
                            "name": "code",
                            "type": "int"
                        }
                    ]
                }
            ],
            "name": "error",
            "default": "None"
        }
    ]
}

payload = {
    'id': 1234,
    'msg_title': "",
    'error': {'message': 'test', 'code': 1},
}

fp = io.BytesIO()
fastavro.schemaless_writer(fp, fastavro.parse_schema(schema), payload)

# let's see how serialization works fine
fp.getvalue()

# '{"id": 1234, "msg_title": "Hi", "date_sent": ('ErrorInfo', {'message': 'test', 'code': 1})}'

# let's go to the initial position on fp
fp.seek(0)

# deserialize
record = fastavro.schemaless_reader(fp, fastavro.parse_schema(schema), return_record_name=True)

print(record)
```
I expected that for this simple case (only null or errorInfo) the reader did not return a tuple but the actual fields or null
Is this an expected return ?

I am trying to deserialize with pydantic and it complexify very much.

untereiner:
Yes, my bad. My question isn't clear. 
I expected this return `{'id': 1234, 'msg_title': '', 'error': {'message': 'test', 'code': 1}}` since the other branch of the union is null. So it is easy to discriminate compared to unions with to records.

scottbelden:
Oh, so is the problem that you have a large schema where there could be a union of `Name1` and `null` and other parts where there could be a union of `AnotherName1` and `AnotherName2`. So in the first case with the `null` you don't want the tuple because it's unambiguous whereas in the second case with multiple named types it is useful to have the tuple?

untereiner:
Ha exactly thank you, for reformulating. Is it possible ? Or does it break the avro serialisation/deserialisation spec ?

scottbelden:
[...]

PR #626, opened 2022-07-13
feat: handle the case of simple unions

untereiner:
ADD the possibility to return only the type when the union is explicit
Issue #625

### DIFF
--- a/fastavro/_validation_py.py
+++ b/fastavro/_validation_py.py
@@ -331,6 +331,7 @@ VALIDATORS = {
     "map": validate_map,
     "union": validate_union,
     "error_union": validate_union,
+    "nullable_union": validate_union,
     "record": validate_record,
     "error": validate_record,
     "request": validate_record,
--- a/fastavro/_validation_py.py
+++ b/fastavro/_validation_py.py
@@ -331,7 +331,6 @@ VALIDATORS = {
     "map": validate_map,
     "union": validate_union,
     "error_union": validate_union,
-    "nullable_union": validate_union,
     "record": validate_record,
     "error": validate_record,
     "request": validate_record,
--- a/fastavro/utils.py
+++ b/fastavro/utils.py
@@ -60,7 +60,11 @@ def gen_data(schema: Schema, named_schemas: NamedSchemas, index: int) -> Any:
             _gen_utf8(): gen_data(map_schema["values"], named_schemas, index)
             for _ in range(10)
         }
-    elif record_type == "union" or record_type == "error_union":
+    elif (
+        record_type == "union"
+        or record_type == "error_union"
+        or record_type == "nullable_union"
+    ):
         union_schema = cast(List[Schema], schema)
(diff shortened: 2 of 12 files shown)


fastavro/_validation_py.py
```diff
@@ -331,6 +331,7 @@ def validate_union(datum, schema, named_schemas, parent_ns=None, raise_errors=Tr
     "map": validate_map,
     "union": validate_union,
     "error_union": validate_union,
+    "nullable_union": validate_union,
```
scottbelden, line 334: This isn't needed anymore, right?

fastavro/utils.py
```diff
@@ -60,7 +60,11 @@ def gen_data(schema: Schema, named_schemas: NamedSchemas, index: int) -> Any:
             _gen_utf8(): gen_data(map_schema["values"], named_schemas, index)
             for _ in range(10)
         }
-    elif record_type == "union" or record_type == "error_union":
```
scottbelden, line 63: This can be reverted back, right?

scottbelden (changes requested):

tests/test_fastavro.py
```diff
@@ -1985,6 +1985,43 @@ def test_return_record_name_with_named_type_in_union():
     assert records == rt_records
 
 
+def test_return_record_with_named_type_in_union():
```
scottbelden, line 1988: Can we add another test where there is a union of `null`, `record1`, and `record2` and someone sets `return_record_name` to `True`? The code as is looks like it would consider that a nullable union and override the `return_record_name` which we wouldn't want.

scottbelden:
The intention here is to only handle simple unions (union of null and another record), right? It feels like we could extend this to be a union of any size as long as there is only one record, but perhaps that's best left out for now since I believe it should be possible to expand to that without a breaking change.

fastavro/_validation_py.py
untereiner: yes, my mistake

fastavro/utils.py
untereiner: yep

tests/test_fastavro.py
untereiner: I believe you are right. I could add a len == 2 in the is_nullable test

untereiner:
It is the exact intent. I don't know yet if it is extendable to unions of any size of simple types + one record. I try something

untereiner:
I juste have a doubt about the test: `if extracted_type not in AVRO_TYPES or extracted_type == "record":`
The "record" type is also in AVRO_TYPES. 

Should AVRO_TYPES not be separated into primitives types and complex types ? 
Knowing that this simple union thing works only with primitive types I guess ?

untereiner:
I didn't think so far. I thought it wouldn't work. But yes if it works I would expect to get the value and not a tuple.
For me it is all good! 
Thanks for working on it !

PR #650, opened 2022-10-13
add default value checks when parsing the schema

scottbelden:
Resolves https://github.com/fastavro/fastavro/issues/649


scottbelden:
Python implementation done. Need to clean it up and then update the cython files.

PR #655, opened 2022-11-02
Improve error messages

retoo:
Better error messages when an enum contains an unknown item or when reader schema doesn't have a default set for a missing value.

**Before**

```
fastavro._read_common.SchemaResolutionError: CLUBS not found in reader symbol list ['SPADES', 'HEARTS', 'DIAMONDS']
```

**After**
```
fastavro._read_common.SchemaResolutionError: CLUBS not found in reader symbol list test.Suit, known symbols: ['SPADES', 'HEARTS', 'DIAMONDS']
```

@scottbelden the new error messages are just an suggestion, they can be formatted differently. 

What do you think?

### DIFF
--- a/fastavro/_read.pyx
+++ b/fastavro/_read.pyx
@@ -322,7 +322,7 @@ cpdef read_enum(fo, writer_schema, reader_schema):
             return default
         else:
             symlist = reader_schema["symbols"]
-            msg = f"{symbol} not found in reader symbol list {symlist}"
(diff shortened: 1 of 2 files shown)


retoo:
@scottbelden what about the _py.py files, should we add the messages there as well?

scottbelden:
This sounds good to me.

> @scottbelden what about the _py.py files, should we add the messages there as well?

Yeah, you will need to update those as well.

scottbelden:
Merged via https://github.com/fastavro/fastavro/pull/717

Issue #676, opened 2023-02-14
Be more forgiving of incorrect schemas in Avro Object Container Files


spenczar:
I have a few terabytes of scientific data in Avro Object Container Files. The files were written with an old version of fastavro which was forgiving of schema issues. This data has a schema which is no longer permitted after https://github.com/fastavro/fastavro/pull/650. It has fields like this, which have a `default` which doesn't match the first type in their union:

```python
     {'default': 0,
      'doc': '1 => candidate is from a Target-of-Opportunity (ToO) exposure; 0 => candidate is from a non-ToO exposure',
      'name': 'tooflag',
      'type': ['null', 'int']}
},
```

This schema is certainly incorrect, which is regrettable. But I still need to read all that data. 

I have constructed a corrected schema with the type ordering fixed; it passes through `fastavro.schema.parse_schema` without errors, so I believe it's valid.

I try to pass it in to `fastavro.reader`, like `fastavro.reader("path_to_data.avro", reader_schema=corrected_schema)`, but no dice:
```


File fastavro/_read.pyx:1070, in fastavro._read.reader.__init__()

File fastavro/_read.pyx:1044, in fastavro._read.file_reader.__init__()

File fastavro/_schema.pyx:106, in fastavro._schema.parse_schema()

File fastavro/_schema.pyx:322, in fastavro._schema._parse_schema()

File fastavro/_schema.pyx:381, in fastavro._schema.parse_field()

File fastavro/_schema.pyx:322, in fastavro._schema._parse_schema()

File fastavro/_schema.pyx:381, in fastavro._schema.parse_field()

File fastavro/_schema.pyx:151, in fastavro._schema._parse_schema()

File fastavro/_schema.pyx:184, in fastavro._schema._parse_schema()

File fastavro/_schema.pyx:132, in fastavro._schema._raise_default_value_error()

SchemaParseException: Default value <0> must match first schema in union with type: null
```

Note that this is error is coming from fastavro/_read.pyx:1044, which is the line that parses the _writer_ schema: https://github.com/fastavro/fastavro/blob/a1a2d36b0e1d0aaf78c4cd04f82d1f5a43ed01db/fastavro/_read.pyx#L1044

I get that this is against the spec, but it's pretty harsh to throw away 5 years of data! Right now, I can get away with pinning to fastavro==1.6.1, but that feels pretty bad. I'd like it if fastavro could somehow be more forgiving of incorrect writer schemas.

scottbelden:
[...]

PR #677, opened 2023-02-21
allow union default errors when a reader schema is supplied

scottbelden:
Resolves https://github.com/fastavro/fastavro/issues/676


spenczar:
Thanks, this looks like it definitely fixes #676 for me.

PR #704, opened 2023-07-18
Allows referencing mentioned schema by name

SerikDM:
https://github.com/fastavro/fastavro/issues/700

### DIFF
--- a/fastavro/_read.pyx
+++ b/fastavro/_read.pyx
@@ -43,12 +43,18 @@ class ReadError(Exception):
     pass
 
 
-cpdef match_types(writer_type, reader_type):
+cpdef _default_named_schemas():
+    return {"writer": {}, "reader": {}}
+
+
+cpdef match_types(writer_type, reader_type, named_schemas=None):
+    if named_schemas is None:
+        named_schemas = _default_named_schemas()
     if isinstance(writer_type, list) or isinstance(reader_type, list):
         return True
     if isinstance(writer_type, dict) or isinstance(reader_type, dict):
         try:
-            return match_schemas(writer_type, reader_type)
+            return match_schemas(writer_type, reader_type, named_schemas)
         except SchemaResolutionError:
             return False
     if writer_type == reader_type:
@@ -64,10 +70,16 @@ cpdef match_types(writer_type, reader_type):
         return True
     elif writer_type == "bytes" and reader_type == "string":
         return True
+    writer_schema = named_schemas["writer"].get(writer_type)
+    reader_schema = named_schemas["reader"].get(reader_type)
+    if writer_schema is not None and reader_schema is not None:
+        return match_types(writer_schema, reader_schema, named_schemas)
     return False
 
 
-cpdef match_schemas(w_schema, r_schema):
+cpdef match_schemas(w_schema, r_schema, named_schemas=None):
+    if named_schemas is None:
+        named_schemas = _default_named_schemas()
     error_msg = f"Schema mismatch: {w_schema} is not {r_schema}"
     if isinstance(w_schema, list):
         # If the writer is a union, checks will happen in read_union after the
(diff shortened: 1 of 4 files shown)


fastavro/_read.pyx
```diff
@@ -64,10 +70,16 @@ cpdef match_types(writer_type, reader_type):
...
         return True
+    writer_schema = named_schemas["writer"].get(writer_type)
+    reader_schema = named_schemas["reader"].get(reader_type)
+    if writer_schema is not None and reader_schema is not None:
+        return match_types(writer_schema, reader_schema, named_schemas)
     return False
 
 
-cpdef match_schemas(w_schema, r_schema):
+cpdef match_schemas(w_schema, r_schema, named_schemas=None):
```
scottbelden, line 80: I think named_schemas is always passed in, so this can just be a required arg instead of optional.

tests/test_fastavro.py
```diff
@@ -3392,3 +3392,69 @@ def test_allow_bad_default_if_correct_reader_schema_present():
...
+            },
+        ],
+    }
+
+    records = [{"foo": {"spam": "ham"}, "bar": [{"spam": "ham"}]}]
+
+    # Parse the schema and ignore union default errors like it would have been
+    # parsed in the past
+    parsed_writer_schema = fastavro.parse_schema(
+        writer_schema, _ignore_default_error=True
```
scottbelden, line 3454: Was the `_ignore_default_error` copied from another test? I don't think it applies here.

SerikDM: Yep, copypaste

fastavro/_read.pyx
```diff
@@ -43,12 +43,18 @@ class ReadError(Exception):
     pass
 
 
-cpdef match_types(writer_type, reader_type):
+cpdef _default_named_schemas():
+    return {"writer": {}, "reader": {}}
+
+
+cpdef match_types(writer_type, reader_type, named_schemas):
+    if named_schemas is None:
```
scottbelden, line 51: This check shouldn't be necessary

fastavro/_read.pyx
```diff
@@ -64,10 +70,16 @@ cpdef match_types(writer_type, reader_type):
...
+    writer_schema = named_schemas["writer"].get(writer_type)
+    reader_schema = named_schemas["reader"].get(reader_type)
+    if writer_schema is not None and reader_schema is not None:
+        return match_types(writer_schema, reader_schema, named_schemas)
     return False
 
 
-cpdef match_schemas(w_schema, r_schema):
+cpdef match_schemas(w_schema, r_schema, named_schemas):
+    if named_schemas is None:
```
scottbelden, line 81: This check shouldn't be necessary

fastavro/_read_py.py
```diff
@@ -42,12 +42,18 @@
 epoch_naive = datetime(1970, 1, 1)
 
 
-def match_types(writer_type, reader_type):
+def _default_named_schemas() -> Dict[str, NamedSchemas]:
+    return {"writer": {}, "reader": {}}
+
+
+def match_types(writer_type, reader_type, named_schemas):
+    if named_schemas is None:
```
scottbelden, line 50: This check shouldn't be necessary

fastavro/_read_py.py
```diff
@@ -63,10 +69,16 @@ def match_types(writer_type, reader_type):
...
+    writer_schema = named_schemas["writer"].get(writer_type)
+    reader_schema = named_schemas["reader"].get(reader_type)
+    if writer_schema is not None and reader_schema is not None:
+        return match_types(writer_schema, reader_schema, named_schemas)
     return False
 
 
-def match_schemas(w_schema, r_schema):
+def match_schemas(w_schema, r_schema, named_schemas):
+    if named_schemas is None:
```
scottbelden, line 80: This check shouldn't be necessary

scottbelden (changes requested):
Sorry, just a few small lines that can be removed and then we should be good to merge.

PR #728, opened 2023-10-24
Update enum symbol validation to enforce avro spec on the entire symbol

### LINKED ISSUE #727: `parse_schema` allows enum symbols with spaces
According to the Avro spec for [enum symbols](https://avro.apache.org/docs/1.11.1/specification/#enums:~:text=this%20schema%20(optional).-,symbols,-%3A%20a%20JSON%20array), a symbol is not allowed to contain spaces:
>a JSON array, listing symbols, as JSON strings (required). All symbols in an enum must be unique; duplicates are prohibited. Every symbol must match the regular expression [A-Za-z_][A-Za-z0-9_]* (the same requirement as for [names](https://avro.apache.org/docs/1.11.1/specification/#names)).

This was partially addressed in the resolution to issue [551](https://github.com/fastavro/fastavro/issues/551), but misses some edge cases. For instance, the string symbol in the following example has spaces, but `parse_schema` doesn't raise an exception:
```python
import fastavro

if __name__ == '__main__':
    schema = {
        'type': 'record',
        'name': 'test_record',
          "fields": [
            {
                "name": "test_field",
                "type": {
                    "type": "enum",
                    "name": "Test_Field",
                    "symbols": ['string with spaces']
                }
            }
          ]
    }
    fastavro.parse_schema(schema)
```

This is due to the fact that [`re.match`](https://docs.python.org/3/library/re.html#re.match) is used in the [`_validate_enum_symbols`](https://github.com/fastavro/fastavro/blob/master/fastavro/_schema.pyx#L759) method, and `re.match` only enforces a match on the beginning of the string rather than the entire string. So in this case it matches `'string'` and stops. We can see this play out with a dumbed down example:
```python
import re
pattern = '[A-Za-z_][A-Za-z0-9_]*'
print(re.match(pattern=pattern, string='string with spaces'))
# <re.Match object; span=(0, 6), match='string'>
```
Changing `re.match` to [`re.fullmatch`](https://docs.python.org/3/library/re.html#re.fullmatch) yields the desired behavior:
```python
import re
pattern = '[A-Za-z_][A-Za-z0-9_]*'
print(re.fullmatch(pattern=pattern, string='string with spaces'))
# None
```

OffByOnee:
Resolves #727.

`re.match` would return a match if the beginning of the string matched the regex pattern. This allowed ill-formed symbols to pass validation if the beginning of the string complied with the Avro spec. `re.fullmatch` only returns a match if the entire string matches the regex pattern, which yields behavior in line with the Avro spec.

Additionally, note that invoking the `fullmatch` method from a compiled pattern runs approximately 3.6x faster than invoking `fullmatch` from the `re` package and passing the compiled pattern as a parameter. This is likely negligible in the grand scheme of things, but we might as well opt for the faster approach.

Simple benchmark script used to ascertain those results:
```python
import re
import timeit

if __name__ == '__main__':
    symbol = re.compile(r"[A-Za-z_][A-Za-z0-9_]*")
    s = 'string with spaces'

    def test_re():
        return re.fullmatch(symbol, s)

    def test_symbol():
        return symbol.fullmatch(s)

    assert test_re() == test_symbol()

    R = 50
    r = timeit.repeat(stmt='test_re()', repeat=R, globals=globals())
    print(f're.fullmatch: {min(r)}')

    r = timeit.repeat(stmt='test_symbol()', repeat=R, globals=globals())
    print(f'symbol.fullmatch: {min(r)}')

```
which produced:
```
re.fullmatch: 0.4571467499999926
symbol.fullmatch: 0.1259777919995031
```

### DIFF
--- a/fastavro/_schema.pyx
+++ b/fastavro/_schema.pyx
@@ -756,7 +756,7 @@ def fingerprint(parsing_canonical_form, algorithm):
 def _validate_enum_symbols(schema):
     symbols = schema["symbols"]
     for symbol in symbols:
-        if not isinstance(symbol, str) or not re.match(SYMBOL_REGEX, symbol):
+        if not isinstance(symbol, str) or not SYMBOL_REGEX.fullmatch(symbol):
             raise SchemaParseException(
                 "Every symbol must match the regular expression [A-Za-z_][A-Za-z0-9_]*"
             )
--- a/fastavro/_schema_py.py
+++ b/fastavro/_schema_py.py
@@ -1035,7 +1035,7 @@ def fingerprint(parsing_canonical_form: str, algorithm: str) -> str:
 def _validate_enum_symbols(schema):
     symbols = schema["symbols"]
     for symbol in symbols:
-        if not isinstance(symbol, str) or not re.match(SYMBOL_REGEX, symbol):
+        if not isinstance(symbol, str) or not SYMBOL_REGEX.fullmatch(symbol):
             raise SchemaParseException(
                 "Every symbol must match the regular expression [A-Za-z_][A-Za-z0-9_]*"
             )
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1110,7 +1110,7 @@ def test_explicit_null_namespace_2():
     assert parsed_schema["name"] == "my_schema"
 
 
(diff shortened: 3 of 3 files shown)


scottbelden:
Thanks!

PR #738, opened 2023-12-20
Allow arbitrary ordering of union fields

### LINKED ISSUE #723: Loosen restrictions on default value; it no longer has to match the first schema
@scottbelden heads up, I think this has been fixed in avro, and it will be released in java for `0.12.0`. Rust already has this fix on `0.16.0`.

**resources**
- https://issues.apache.org/jira/browse/AVRO-3649
- https://github.com/apache/avro/pull/2503
- [yet to be released docs](https://github.com/apache/avro/blob/main/doc/content/en/docs/%2B%2Bversion%2B%2B/Specification/_index.md#:~:text=(Note%20that%20when%20a%20%5Bdefault%20value%5D(%7B%7B%3C%20ref%20%22%23schema%2Drecord%22%20%3E%7D%7D%20%22Schema%20record%22)%20is%20specified%20for%20a%20record%20field%20whose%20type%20is%20a%20union%2C%20the%20type%20of%20the%20default%20value%20must%20match%20with%20one%20element%20of%20the%20union)

_Originally posted by @woile in https://github.com/fastavro/fastavro/issues/649#issuecomment-1757080342_

jvansanten:
Since the Java reference implementation no longer enforces the requirement that the default value be the first type in a union, drop this requirement from fastavro as well.  This restores the ability to read data written with fastavro <1.7, while still checking that the default is (or is convertible to) one of the union types. Along the way, refactor the special cases for double and float so that they do not raise exceptions.

Fixes #723

### DIFF
--- a/fastavro/_schema.pyx
+++ b/fastavro/_schema.pyx
@@ -183,11 +183,11 @@ def parse_schema(
 
 
 cdef _raise_default_value_error(
-    default, schema_type, in_union, ignore_default_error
+    default, schema_type, ignore_default_error
 ):
     if ignore_default_error:
         return
-    elif in_union:
+    elif isinstance(schema_type, list):
         text = f"a schema in union with type: {schema_type}"
     else:
         text = f"schema type: {schema_type}"
@@ -226,7 +226,6 @@ cdef _parse_schema(
     named_schemas,
     default,
     ignore_default_error,
-    in_union=False,
 ):
     # union schemas
     if isinstance(schema, list):
@@ -248,7 +247,7 @@ cdef _parse_schema(
                 if _default_matches_schema(default, s):
                     break
             else:
-                _raise_default_value_error(default, schema, True, ignore_default_error)
+                _raise_default_value_error(default, schema, ignore_default_error)
         return parsed_schemas
 
     # string schemas; this could be either a named schema or a primitive type
@@ -257,7 +256,7 @@ cdef _parse_schema(
             if default is not NO_DEFAULT:
                 if not _default_matches_schema(default, schema):
                     _raise_default_value_error(
-                        default, schema, in_union, ignore_default_error
+                        default, schema, ignore_default_error
                     )
             return schema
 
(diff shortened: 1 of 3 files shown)


fastavro/_schema.pyx
```diff
@@ -208,51 +230,32 @@ cdef _parse_schema(
...
-                        namespace,
-                        expand,
-                        False,
-                        names,
-                        named_schemas,
-                        NO_DEFAULT,
-                        ignore_default_error,
-                    )
-                )
+                _raise_default_value_error(default, schema, True, ignore_default_error)
```
scottbelden, line 251: Can you change this to

```
_raise_default_value_error(default, schema, in_union=True, ignore_default_error=ignore_default_error)
```

So that it's more clear what the `True` is for

fastavro/_schema.pyx
```diff
@@ -208,51 +230,32 @@ cdef _parse_schema(
...
-                    _parse_schema(
-                        s,
-                        namespace,
-                        expand,
-                        False,
-                        names,
-                        named_schemas,
-                        default,
-                        ignore_default_error,
-                        in_union=True,
```
scottbelden, line 224: I think we can remove `in_union` from all function calls except for `_raise_default_value_error`, right? Previously it needed to be propagated everywhere because the default wasn't checked right away, but with these changes I don't think we need it in all the function signatures anymore.

scottbelden (changes requested):
Just a few small changes and we should be good to merge. I only made comments on the pyx file, but the same comments apply to the py file as well.

scottbelden:
I had a few questions/comments, but in general this looks great. Thanks so much for the PR!

fastavro/_schema.pyx
jvansanten: I went ahead and removed in_union entirely, since the only thing it was used for was switching the error message in _raise_default_value_error. That can also be done by checking whether the schema arg is a list.

fastavro/_schema.pyx
jvansanten: Resolved by removing is_union entirely.

scottbelden (approved):

scottbelden:
Thanks again! I'll try to get a new release out shortly.

PR #741, opened 2024-01-05
consider the logical type when generating random data

hwse:
The utility functions utils.generate_many and utils.generate_one currently do not consider the logical types. This means for a `string` field with logical type `uuid` invalid values may be generated.

fastavro will then refuse to encode these values. This prevents users to generate random avro records if their schema contains `uuid` fields. To fix this i added a test and basic implementation for all logical types supported by fastavro.

### DIFF
--- a/fastavro/utils.py
+++ b/fastavro/utils.py
@@ -1,12 +1,28 @@
+import datetime
+import uuid
 from hashlib import md5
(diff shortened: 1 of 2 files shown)


scottbelden:
Thanks for the improvement!

PR #742, opened 2024-01-08
test with zlibng

scottbelden:
Resolves https://github.com/fastavro/fastavro/issues/740 and https://github.com/fastavro/fastavro/issues/605

### DIFF
--- a/developer_requirements.txt
+++ b/developer_requirements.txt
@@ -16,3 +16,4 @@ cramjam
 # Also don't install on windows for similar reasons
 zstandard; sys_platform != 'win32'
 lz4; sys_platform != 'win32'
+zlib_ng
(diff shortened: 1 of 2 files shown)


musicinmybrain:
I used the proposed `python-cramjam` package I’m working on for Fedora (pending a few package reviews), and a `python-zlib-ng` package I just put together, to test local builds in Fedora Rawhide (future Fedora 40), 39, and 38.

Fedora Rawhide/40 has the zlib-compat interfaces of zlib-ng as the system-wide zlib implementation (including for the Python standard library), and the others have zlib-ng packaged, but not replacing classic zlib.

At least on `x86_64`, this is working for me in all three releases! Thank you.

scottbelden:
Awesome. Thanks for the original detailed report.

PR #749, opened 2024-02-09
accept alias without namespace

monika-gz:
solves this issue:
https://github.com/fastavro/fastavro/issues/648


scottbelden:
Thanks. Can you add the test case from the issue? You can probably find similar tests in tests/test_fastavro.py that use a reader and writer schema and do a roundtrip.

scottbelden:
Merged via https://github.com/fastavro/fastavro/pull/751

PR #750, opened 2024-02-11
style: simplify condition in _parse_schema method

manawasp:
Hey,

I was working on fixing [python-schema-registry-client](https://github.com/marcosschroh/python-schema-registry-client) bump to fastavro 1.9.3 and due typing issue I end up here reading the code source, I found the following code block can be simplified.

Code behavior should not change via this PR.

_I read the Contributing part and I think this change could go straight to a PR. If you mind differently don't hesitate to tell me_

### DIFF
--- a/fastavro/_schema.pyx
+++ b/fastavro/_schema.pyx
@@ -72,7 +72,7 @@ cpdef inline extract_record_type(schema):
     if isinstance(schema, list):
         return "union"
 
-    return schema  
+    return schema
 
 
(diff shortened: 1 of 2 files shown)


scottbelden:
Thanks for simplifying this. Would you mind also updating the _schema.pyx (https://github.com/fastavro/fastavro/blob/master/fastavro/_schema.pyx#L268-L279) file to match?

manawasp:
Sure ! Done + rebase

PR #751, opened 2024-02-12
Fix aliases in schema resolution

hwse:
Same as https://github.com/fastavro/fastavro/pull/749 but with additional tests. This should fix https://github.com/fastavro/fastavro/issues/648 (added test similar to the example).


scottbelden:
Thanks for the contribution!

PR #753, opened 2024-02-15
fix return_record_name option for aliases in a union

hwse:
These changes should fix a issue with the `return_record_name` and `return_named_type` reader options that occur during schema resolution in a union with aliases.

The issue is that if:

- `return_record_name` is set to True
- or `return_named_type` is set to True

one would expect that the returned names are the names from the reader schema. But this is currently not the case when:

- the writer schema has a union where one of the branches is a named record
- the reader schema changes the name of that record and also adds a alias for the old name

**Then the record name of writer schema is returned**. 

This causes some issues like:

- no being able to re-encode the message with the reader schema (the reader schema has no record with the old name)
- having to check the old and new name in code

### DIFF
--- a/fastavro/_read.pyx
+++ b/fastavro/_read.pyx
@@ -525,6 +525,7 @@ cpdef read_union(
     # schema resolution
     index = read_long(fo)
     idx_schema = writer_schema[index]
+    idx_reader_schema = None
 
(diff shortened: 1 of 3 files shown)


scottbelden:
This looks great. Thanks!

PR #770, opened 2024-05-17
Patch boolean conversion

### LINKED ISSUE #769: `json_writer` and `json_reader` do not respect `boolean` types assigned with `int` values
When presented with a boolean field in the schema, but the data object has an integer instead of a boolean for this field, the JSON reader and writer do nothing to try to transform the data into a boolean.  The binary writers do transform an integer into a boolean type before encoding the message.

I've created a simple example:

```
#!/bin/env python

from io import StringIO
import fastavro
import json

SCHEMA = """
{
    "namespace": "com.example",
    "type": "record",
    "name": "BooleanTest",
    "fields": [
        {
            "name": "boolean",
            "type": "boolean"
        }
    ]
}"""


class JsonAvroEncoder():
  def __init__(self, schema):
    self._schema_str = schema

  @property
  def _schema(self):
    schema_obj = json.loads(self._schema_str)
    return fastavro.parse_schema(schema_obj)

  def encode(self, data_obj) -> str:
    output = StringIO()
    fastavro.json_writer(output, self._schema, [data_obj])
    return output.getvalue()

  def decode(self, message):
    return list(fastavro.json_reader(StringIO(message), self._schema))[0]


if __name__ == '__main__':
  encoder = JsonAvroEncoder(SCHEMA)
  message = { 'boolean': 5 }

  print("Original Message")
  print(message)
  print()

  encoded = encoder.encode(message)
  print("Encoded Message")
  print(encoded)
  print()

  decoded = encoder.decode(encoded)
  print("Decoded Message")
  print(decoded)
  print()
```

This results int he following output:

```
Original Message
{'boolean': 5}

Encoded Message
{"boolean": 5}

Decoded Message
{'boolean': 5}
```

I would expect the `json_writer` to either reject the value given for the `boolean` field or (like the binary writers do) change the `5` into a `True` value like Python does.  But instead, the JSON generated contains the value `5` for the field.  Then the `json_reader` finds an integer `5` in the encoded message and instead of creating a `boolean` like the schema dictates, it creates the integer `5` in the decoded object.  I can't say what the binary readers would do if they found an integer `5` in the encoded data as I haven't tried due to needing to edit the binary by hand.

dasunpubudumal:
Proposed to fix #769

### DIFF
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -129,6 +129,20 @@ def test_json():
     assert records == new_records
 
 
+def test_boolean_conversion():
+    schema = {
+        "type": "record",
+        "name": "test_boolean_conversion",
+        "namespace": "test",
+        "fields": [
+            {"name": "boolean", "type": "boolean"},
+        ],
+    }
+    records = [{"boolean": 1}, {"boolean": 10}, {"boolean": 0}]
+    converted_records = [{"boolean": True}, {"boolean": True}, {"boolean": False}]
+    assert converted_records == roundtrip(schema, records)
+
+
 def test_more_than_one_record():
     schema = {
         "type": "record",
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -143,6 +143,31 @@ def test_boolean_conversion():
     assert converted_records == roundtrip(schema, records)
 
 
+def test_boolean_conversion_with_nulls():
+    schema = {
+        "type": "record",
+        "name": "test_boolean_conversion_with_nulls",
+        "namespace": "test",
+        "fields": [
+            {
+                "name": "boolean",
+                "type": [
+                    "null",
+                    "boolean",
+                ],
+            },
+        ],
+    }
+    records = [{"boolean": 1}, {"boolean": None}, {"boolean": True}, {"boolean": 10}]
+    converted_records = [
+        {"boolean": True},
+        {"boolean": None},
+        {"boolean": True},
+        {"boolean": True},
+    ]
+    assert converted_records == roundtrip(schema, records)
+
+
 def test_more_than_one_record():
     schema = {
         "type": "record",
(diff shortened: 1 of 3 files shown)


tests/test_json.py
```diff
@@ -129,6 +129,20 @@ def test_json():
...
+        "namespace": "test",
+        "fields": [
+            {"name": "boolean", "type": "boolean"},
+        ],
+    }
+    records = [{"boolean": 1}, {"boolean": 10}, {"boolean": 0}]
+    converted_records = [{"boolean": True}, {"boolean": True}, {"boolean": False}]
+    assert converted_records == roundtrip(schema, records)
+
+
```
yoldas, line 145: Although this shows the type conversion, it does not show what the serialised representation is and how the other side reads it though. If it is JSON, it needs to be true/false written. If it is binary, it needs to be a single byte 0/1 written according to the spec. The other end needs a consistent read as well. One would expect the Avro union data types work properly too .

dasunpubudumal: The change was made for `AvroJSONEncoder`. Presumably it doesn't affect binary serialisations?

tests/test_json.py
```diff
@@ -129,6 +129,45 @@ def test_json():
...
+def test_boolean_conversion_with_nulls():
```
dasunpubudumal, line 146: @yoldas check this.

dasunpubudumal:
Notes:

- The `_validate` function at `fastavro/_write_py.py:187` is called upon when `AvroJSONEncoder` calls the `write` function.
- The trouble is that this `_validate` function is common for both binary and JSON encoders.
- If the `_validate` function is updated in `fastavro/_write_py.py:187`, and we only update the `AvroJSONEncoder` to match boolean encoding and decoding, the binary encoder will throw.

PR #784, opened 2024-07-19
Fix deterministic output for gen_many under certain conditions

hwse:
When using the `gen_many` function in `fastavro.utils` one would expect that every combination of enum values or union branches is generated at some point, if a lot of message are generated.

But for some schemas the function returns deterministic output. This seems to be caused by the usage of a incrementing index that is currently used to generate booleans, enums and union branches.

A example schema:

```
record MessageBranchA
{
    string? optional_name;
}

enum SomeEnum
{
    A, B
}

record MessageBranchB
{
    SomeEnum e;
}

record Message
{
    union {MessageBranchA, MessageBranchB} main_branch;
}
```

When generating random Message values, we will see:

* index 0 (%2 = 0): select MessageBranchA, set optional_name to null ( string? <=> union {null,string})
* index 1 (%2 = 1): select MessageBranchB, set enum to B
* index 2 (%2 = 0): select MessageBranchA, set optional_name to null ( string? <=> union {null,string})
* index 1 (%2 = 1): select MessageBranchB, set enum to B

Can be reproduced with this script:

```python
from fastavro.utils import generate_many

schema = {
    "name": "Message",
    "type": "record",
    "fields": [
        {
            "name": "main_branch",
            "type": [
                {
                    "name": "MessageBranchA",
                    "type": "record",
                    "fields": [
                        {"name": "optional_name", "type": ["null", "string"]}
                    ],
                },
                {
                    "name": "MessageBranchB",
                    "type": "record",
                    "fields": [
                        {
                            "name": "e",
                            "type": {
                                "type": "enum",
                                "name": "enum_field",
                                "symbols": ["A", "B"],
                            },
                        }
                    ],
                },
            ],
        }
    ],
}
for m in generate_many(schema, 10):
    print(m)
```

output:

```
{'main_branch': {'optional_name': None}}
{'main_branch': {'e': 'B'}}
{'main_branch': {'optional_name': None}}
{'main_branch': {'e': 'B'}}
{'main_branch': {'optional_name': None}}
{'main_branch': {'e': 'B'}}
{'main_branch': {'optional_name': None}}
{'main_branch': {'e': 'B'}}
{'main_branch': {'optional_name': None}}
{'main_branch': {'e': 'B'}}
```

I removed the index parameter and used `random.randint` instead. Added no unit test since testing that something is random is hard.

### DIFF
--- a/fastavro/utils.py
+++ b/fastavro/utils.py
@@ -38,7 +38,7 @@ def _gen_utf8() -> str:
     return "".join(random.choices(ascii_letters, k=10))
 
 
-def gen_data(schema: Schema, named_schemas: NamedSchemas, index: int) -> Any:
+def gen_data(schema: Schema, named_schemas: NamedSchemas) -> Any:
     record_type = extract_record_type(schema)
     logical_type = extract_logical_type(schema)
 
@@ -77,7 +77,7 @@ def gen_data(schema: Schema, named_schemas: NamedSchemas, index: int) -> Any:
     elif record_type == "double":
         return random.random()
     elif record_type == "boolean":
-        return index % 2 == 0
+        return bool(random.randint(0, 1))
     elif record_type == "bytes":
         return _randbytes(10)
     elif record_type == "fixed":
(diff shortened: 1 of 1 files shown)


scottbelden:
Thanks. I agree that this is probably better.

PR #792, opened 2024-08-27
Resolve referenced schema to disambiguate union types with named schemas

enifsieus:
When a union is handed a named schema reference to write, it does not resolve named schemas, and may use an empty set to disambiguate, causing it to select the first type rather than the correct type.

### DIFF
--- a/fastavro/_write.pyx
+++ b/fastavro/_write.pyx
@@ -308,6 +308,12 @@ cdef write_union(bytearray fo, datum, schema, dict named_schemas, fname, dict op
                 options=options,
             ):
                 record_type = extract_record_type(candidate)
+
+                if record_type in named_schemas:
+                    # Convert named record types into their full schema so that we can check most_fields
+                    candidate = named_schemas[record_type]
+                    record_type = "record"
+
                 if record_type == "record":
                     logical_type = extract_logical_type(candidate)
                     if logical_type:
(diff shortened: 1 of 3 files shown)


enifsieus:
Hello -
Do I need to open an Issue to tie this to and/or explain anything else about what I uncovered, or is the PR itself sufficient? Checking whether the candidate schema is type(str) feels a bit flimsy, but it does the job and I don't know enough about internals to consider alternate approaches.  I welcome feedback here, and hope to help resolve this issue. 

Cheers -
~tmw

scottbelden:
The PR is sufficient and the test case perfectly explains the problem. Thank you for that.

Life is just so busy that I haven't had much time for this project, but this looks simple enough and we should be able to get it merged in shortly.

The implementation seems fine, but I think it might be slightly cleaner as the following:

```diff
            if _validate(
                datum,
                candidate,
                named_schemas,
                raise_errors=False,
                field="",
                options=options,
             ):
                 record_type = extract_record_type(candidate)
+                if record_type in named_schemas:
+                    # Convert named record types into their full schema so that we can check most_fields
+                    candidate = named_schemas[record_type]
+                    record_type = "record"
+
                 if record_type == "record":
                     logical_type = extract_logical_type(candidate)
                     if logical_type:
```

scottbelden:
Also, you will need to modify `_write.pyx` with the same change that is made in `_write_py.py`

enifsieus:
Thank you - I've made the changes requested and pushed an updated changeset!

PR #801, opened 2024-09-15
Issue 794: Print detailed error messages for out-of-range values when validate int or long type. 

jungbumwoo:
## Title
print detailed error messages for out-of-range values when validate int or long type. 

## Changes
- Error messages now include the field name and the problematic value to give more context to the user.
- Introduced ValueError and OverflowError to signal when datum is too small or too large, respectively.
- The function now raises:
ValueError if datum is smaller than INT_MIN_VALUE, LONG_MIN_VALUE.
OverflowError if datum exceeds INT_MAX_VALUE, LONG_MAX_VALUE.
- bool values are explicitly checked and rejected to avoid being treated as integers before range validation

## Related issues
https://github.com/fastavro/fastavro/issues/794

### DIFF
--- a/fastavro/_validate_common.py
+++ b/fastavro/_validate_common.py
@@ -20,3 +20,16 @@ class ValidationError(Exception):
         message = json.dumps([str(e) for e in errors], indent=2, ensure_ascii=False)
         super().__init__(message)
         self.errors = errors
+
+
+class ValidationValueErrorData(
+    namedtuple("ValidationValueErrorData", ["datum", "limit", "field"])
+):
+    def __str__(self):
+        return f"{self.field} <{self.datum}> out of range {self.limit}"
+
+
+class ValidationValueError(ValueError):
+    def __init__(self, error, *args, **kwargs):
+        message = json.dumps(str(error), indent=2, ensure_ascii=False)
+        super().__init__(message, *args, **kwargs)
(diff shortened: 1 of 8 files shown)


scottbelden:
Can you add a test case to show that the new error message shows up?

jungbumwoo:
@scottbelden Hello, I apologize for the delayed response. I've been quite busy with work recently. As per your suggestion, I have added the test code. Could you please take a look when you have a moment? I would also appreciate any additional feedback on the PR or anything else. Thank you!

jungbumwoo:
By the way, how can I run the test code in this PR?

scottbelden:
No worries. I also have been slammed with work. I'll try to take a look at this and see if there are any suggestions.

jungbumwoo:
@scottbelden scott, build was failed at pypy3.9. so merge upstream to apply this commit [this commit](https://github.com/fastavro/fastavro/commit/cdf1d995f737a8100f7f46c970fb958cd96ae015) : )

scottbelden:
I noticed if I did something like `validate(INT_MAX_VALUE + 1, "int", "foo", raise_errors=False)` it would still raise an error about the value being larger than the maximum size for an int. I have some local changes that fix this, but it's not quite fully working. I'm going to try to get it fully working tomorrow and push it.

PR #846, opened 2025-07-04
Improve performance for schema resolution

hwse:
I observered a big slowdown when using a different writer and reader schema compared to writer schema == reader schema. 

This can be reproduced with this script:

```python
import io
import random
import time
from contextlib import contextmanager

import fastavro
import fastavro.utils

print(fastavro)


def generate_big_schema(field_count):
    def field_nr(nr: int):
        return {
            "name": f"field_{nr}",
            "type": "int",
            "default": field_count,
            "doc": "Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam"
        }

    return {
        "type": "record",
        "name": "BigSchema",
        "fields": [field_nr(nr) for nr in range(field_count)]
    }


WRITER_SCHEMA = fastavro.schema.parse_schema(generate_big_schema(100))
READER_SCHEMA = fastavro.schema.parse_schema(generate_big_schema(101))


def serialize(msg):
    buffer = io.BytesIO()
    fastavro.schemaless_writer(buffer, WRITER_SCHEMA, msg)
    return buffer.getvalue()


@contextmanager
def timer(name):
    start = time.perf_counter()
    yield
    stop = time.perf_counter()
    print(f"{name}: {stop - start}s")


random.seed("foobar")
raw_message = serialize({})

measure = True
if measure:
    ITERATIONS = 100000
    with timer("no resolution"):
        for _ in range(ITERATIONS):
            fastavro.schemaless_reader(io.BytesIO(raw_message), WRITER_SCHEMA, reader_schema=None)

    with timer("resolution"):
        for _ in range(ITERATIONS):
            fastavro.schemaless_reader(io.BytesIO(raw_message), WRITER_SCHEMA, reader_schema=READER_SCHEMA)
else:
    ITERATIONS = 1000
    for _ in range(ITERATIONS):
        fastavro.schemaless_reader(io.BytesIO(raw_message), WRITER_SCHEMA, reader_schema=READER_SCHEMA)
```

on my machine it prints:

```
$ python resolution_bench.py 
<module 'fastavro' from '/tmp/venv/lib/python3.10/site-packages/fastavro/__init__.py'>
no resolution: 0.6873706740007037s
resolution: 22.573060028997133s
```

I investigated this a bit using valgrind(callgrind) and saw that the following function took a lot of time:

- PyObject_Repr
- PySequence_Contains

To improve this i changed:

- only build error messages on demand. in some cases the schema was formatted into a string, which can be expansive for big schemas (e.g. big doc string as in the script)
- use a set instead of a list where appropriate

with these changes i now get:

```
no resolution: 0.9697164819954196s
resolution: 3.084980943996925s
```

The ~3x slowdown is still not optimal i guess, but better than before

### DIFF
--- a/fastavro/_read.pyx
+++ b/fastavro/_read.pyx
@@ -82,7 +82,6 @@ cpdef match_types(writer_type, reader_type, named_schemas):
 
 
 cpdef match_schemas(w_schema, r_schema, named_schemas):
-    error_msg = f"Schema mismatch: {w_schema} is not {r_schema}"
     if isinstance(w_schema, list):
         # If the writer is a union, checks will happen in read_union after the
         # correct schema is known
@@ -94,7 +93,9 @@ cpdef match_schemas(w_schema, r_schema, named_schemas):
             if match_types(w_schema, schema, named_schemas):
                 return schema
         else:
-            raise SchemaResolutionError(error_msg)
+            raise SchemaResolutionError(
+                f"Schema mismatch: {w_schema} is not {r_schema}"
+            )
     else:
         # Check for dicts as primitive types are just strings
         if isinstance(w_schema, dict):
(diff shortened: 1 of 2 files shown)


sral85:
@hwse This merge request is also relevant for me. I would really appreciate, if you find the time to have a look into it.

scottbelden:
Sorry for the delay on this. New release should be out today.

hwse:
> Sorry for the delay on this. New release should be out today.

No problem, thanks for merging

PR #860, opened 2025-09-25
Feature: add ability to enforce logical type constraints at serialization/validation

Aweave15:
### Issue

When the value of a logical type is the same type as the avro native type, and not necessarily coerced into the native python type, it passes validation and is serialized even if that is undesired. This can cause issues during deserialization downstream.

The detail above is not necessarily *the* issue, moreso that there doesn't seem to be a reasonable way for users to be able to perform this extra validation without monkeypatching a large chunk of private code or breaking things elsewhere.

This is directly related to https://github.com/fastavro/fastavro/issues/838 

### Example

For example if a user defines a uuid logical type, and the input into serialization/validation is a badly formatted string, then fast avro will serialize the data no problem and claim it is valid. Only failing down the line when it comes to deserialize the data. Example:

```
schema = {
    "type": "record",
    "name": "example",
    "fields": [
        {
             "name": "id",
             "type": {"type": "string", "logicalType": "uuid"}
        }
    ]
}

record = {"id": "bad-uuid"}

# passes
validated = validate(record, schema)

# successful
buf = io.BytesIO()
fastavro.schemaless_writer(buf, schema, record)
serialized_data = buf.getvalue()

# fails
buf.seek(0)
deserialized = fastavro.schemaless_reader(buf, schema)
```

### Sample Usecase:

Pretty much any usecase where the input data is not in native python types - Imagine an http restful service which takes in a schema identifier, and a generic payload corresponding to the data. The service is a gateway for validating and publishing data from clients. It fetches the avro schema via the identifier from the registry, serializes/validates the payload against the schema, and writes it somewhere.

Since it's JSON input (http rest), the types are not coerced into python native types ie. uuids, datetimes, etc. An example above would allow bad data to be serialized/validated, and then impact deserialization downstream.

A user can patch an existing writer, or create their own to perform this validation ie.

```
original_logical_writer = fastavro.write.LOGICAL_WRITERS["string-uuid"]

def enforce_uuid(data, schema):
    if isinstance(data, str):
        try:
            data = UUID(data)
        except ValueError:
            raise TyperError
    return original_logical_writer(data)

fastavro.write.LOGICAL_WRITERS["string-uuid"] = encode_uuid
```

However, the logical writes are used outside of just writing that specific logical type. For example, evaluating/validating against a candidate within a Union, the logical writers are also applied. So this would break union matching and validation.

ie. The above would break union concepts for something like this 
```
{
    "type": "record",
    "name": "test_validate_union_handles_logical_type_errors",
    "fields": [
        {
            "name": "run_data",
            "type": [
                {
                    "name": "legacy_run",
                    "type": "record",
                    "fields": [
                        {
                            "name": "run_id",
                            "type": {"type": "string", "logicalType": "uuid"}
                        },
                        {
                            "name": "count",
                            "type": "int"
                        }
                    ]
                },
{
                    "name": "run",
                    "type": "record",
                    "fields": [
                        {
                            "name": "run_id",
                            "type": {"type": "string", "logical_type": "uuid"}
                        },
                        {
                            "name": "count",
                            "type": "int"
                        }
                    ]
                }
            ]
        }
    ],
}

```

### Solution

I know there's been discussions wrt logical type validation vs hints etc. However it seems contradictory for fastavro to serialize fine, validate fine, and then error at deserialization. It seems reasonable to allow users a way to be able to opt-in to some more strict validation themselves.

Logically, this should essentially be a no-op for all existing users. It should have 0 impact to any users that chose to not opt-in to the functionality.

It introduces a new fastavro native exception, where users can **choose** to implement or patch their own logic for extra validation and raise in the event it doesn't meet their defined constraints, without breaking other operations such as union matching.

I'm new to working in this library, so please do let me know if im missing anything! I tried to go the least intrusive way to support this feature, however open to any pattern desired if it allows writers/validators to be more strict on logical types, similar to the readers. Otherwise monkeypatching/maintaing a fork would be my last option which I would like to avoid. Happy to add more tests as well, mostly wanted to get feedback quicker before investing too heavily in additional test suite 

Here's the benchmark comparison for before vs after the change (using the checked-in benchmark script):
[Pre-change](https://gist.github.com/Aweave15/bca4b1b9985b29238bed42e3d245fec6)
[Post-change](https://gist.github.com/Aweave15/c4c5a82b7c4e55e0eb19a6a8c4ad24e7)


fastavro/_validation.pyx
```diff
@@ -223,7 +223,10 @@ cpdef _validate(
     if logical_type:
         prepare = LOGICAL_WRITERS.get(logical_type)
         if prepare:
-            datum = prepare(datum, schema)
+            try:
+                datum = prepare(datum, schema)
+            except LogicalTypeValidationError:
```
Aweave15, line 228: realized  I need to update this to raise if configured to raise on error - will update that. Interim feedback on the approach still welcomed in the meantime please!

Aweave15:
> I haven't looked over this yet, but did you intend to close it?

My apologies @scottbelden , I closed it temporarily to do some more testing and a couple minor changes. Was a bit busy over the weekend so I closed it until had some time freed up. Reopened, whenever you could take a look!

PR #871, opened 2026-03-30
Add Windows ARM64 support in CI

Harishmcw:
This PR adds Windows ARM64 support in the `build_windows.yml`

- Add `windows-11-arm` as a second OS runner in the build matrix.
- Excludes Python 3.9, 3.10, and PyPy for `windows-11-arm` as they are not supported on Windows ARM64.
- Guards release creation to `windows-latest` only to avoid duplicate GitHub releases on tag push.
- Updates artifact naming to distinguish `x64` vs `arm64` wheels.

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -11,10 +11,13 @@ on:
 
 jobs:
   test:
-    name: test ${{ matrix.py }} - windows
-    runs-on: windows-latest
+    name: test ${{ matrix.py }} - ${{ matrix.os }} 
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


scottbelden:
Thanks! I'll probably try to create a new release so that these wheels are available.

Harishmcw:
> Thanks! I'll probably try to create a new release so that these wheels are available.

Hi @scottbelden,

Could you please let me know when the release will be made? Looking forward to using the wheels via PyPI.

scottbelden:
Sorry about that, this completely slipped my mind. I believe it should be available under version `1.12.2`

PR #875, opened 2026-05-11
add cython hint to declare free threading

### LINKED ISSUE #873: free-threading support
[free threading](https://py-free-threading.github.io) is generally available however this library does not support it

```
$ python
Python 3.14.2 free-threading build (main, Dec  5 2025, 21:11:07) [Clang 21.1.4 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import fastavro
<frozen importlib._bootstrap>:491: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'fastavro._read', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

scottbelden:
Resolves https://github.com/fastavro/fastavro/issues/873

### DIFF
--- a/fastavro/_logical_readers.pyx
+++ b/fastavro/_logical_readers.pyx
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 
 from datetime import datetime, time, date, timezone, timedelta
 from decimal import Context
--- a/fastavro/_schema.pyx
+++ b/fastavro/_schema.pyx
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 
 import hashlib
 from io import StringIO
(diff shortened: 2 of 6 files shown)


fastavro/_logical_readers.pyx
```diff
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
```
ericbuehl, line 2: is the module-level `decimal_context` thread-safe?

fastavro/_schema.pyx
```diff
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
```
ericbuehl, line 2: the `named_schemas` dict also mutated in a few places

fastavro/_write.pyx
```diff
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
```
ericbuehl, line 2: are Writer instances safe to share across threads?

PR #894, opened 2026-08-14
BinaryDecoder: raise EOFError on truncated primitives instead of TypeError/struct.error

eeshsaxena:
`BinaryDecoder.read_bytes` and `read_fixed` already raise `EOFError` when the stream ends early, but the other primitive readers don't, so a truncated stream surfaces as a confusing `TypeError` or `struct.error`:

```python
import io
from fastavro.io.binary_decoder import BinaryDecoder
BinaryDecoder(io.BytesIO(b"\x80")).read_long()
# TypeError: ord() expected a character, but string of length 0 found
```

The cases:

- `read_long` only checks for EOF on the first byte of the varint (its own comment says "We do EOF checking only here"), so a continuation byte at end-of-stream reaches `ord(b"")` and raises `TypeError`.
- `read_boolean`, `read_float` and `read_double` unpack whatever `read()` returns, so a short read raises `struct.error`.

This guards each of them to raise `EOFError` on a short read, matching what `read_bytes`/`read_fixed` already do, so callers can handle a truncated stream the same way everywhere. Valid input is unchanged (verified read_boolean/read_long/read_float/read_double still decode correctly). Added a test.

Scoped to the pure-Python `BinaryDecoder`; I didn't touch the Cython reader path.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`BinaryDecoder: raise EOFError on truncated primitives instead of TypeError/struct.error`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining fastavro!

PR #895, opened 2026-08-14
schema: raise SchemaParseException for a malformed record field

eeshsaxena:
`parse_schema` raises `SchemaParseException` for most invalid schemas, but a malformed record *field* slips through as a bare `AttributeError` or `KeyError`:

```python
import fastavro
fastavro.parse_schema({"type": "record", "name": "r", "fields": "notalist"})
# AttributeError: 'str' object has no attribute 'items'
fastavro.parse_schema({"type": "record", "name": "r", "fields": [{"name": "f"}]})
# KeyError: 'type'
```

`parse_field` assumes each field is a dict with `name` and `type`. A field that isn't a dict (including the case where `fields` is a string, which iterates into characters) hits `field.items()` and raises `AttributeError`, and a field missing `type`/`name` raises `KeyError`.

This validates the field up front and raises `SchemaParseException` with a clear message, consistent with the rest of the parser. The change is applied to both the Python (`_schema_py.py`) and Cython (`_schema.pyx`) parsers so they stay in sync. Added a test; the schema test suite passes (68).


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`schema: raise SchemaParseException for a malformed record field`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining fastavro!

